PR #9, opened 2018-11-19
GLT-2182: report HiSeq and NovaSeq run positions

djcooke:


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
```diff
@@ -89,8 +91,12 @@
       XPathExpression miSeqPartNumber = xpath.compile("//FlowcellRFIDTag/PartNumber/text()");
       XPathExpression nextSeqPartNumber = xpath.compile("//FlowCellRfidTag/PartNumber/text()");
       XPathExpression novaSeqPartNum = xpath.compile("//RfidsInfo/FlowCellMode/text()");
-
       CONTAINER_PARTNUMBER_XPATHS = Collections.unmodifiableSet(Sets.newHashSet(miSeqPartNumber, nextSeqPartNumber, novaSeqPartNum));
+      
+      XPathExpression hiSeqPosition = xpath.compile("//Setup/FCPosition/text()");
```
apmasell: I don't know if this is more or less awful, but all of these XPath expression could be joined:
```
XPathExpression novaSeqPosition = xpath.compile("//Setup/FCPosition/text()|//Side/text()");
```

and then it doesn't have to be a collection.

callunity: I like keeping things separate so we can see which instrument models have which changes

callunity:
Oops sorry for not reviewing this before the release. I spoke with @apmasell and he said he'd release this next Tuesday when he's doing the MISO release, so that things can be properly tested on staging.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
djcooke: Yeah, I think it's more maintainable with them kept separate

PR #31, opened 2019-02-26
Change scheduling policy

apmasell:
In an attempt to avoid weird spikes in the outstanding runs metrics

### DIFF
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
@@ -449,7 +449,7 @@ public class Scheduler {
   public synchronized void start() {
     if (scanDirectoriesFuture == null) {
       scanDirectoriesFuture =
-          scheduler.scheduleAtFixedRate(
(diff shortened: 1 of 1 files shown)


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
```diff
@@ -449,7 +449,7 @@ public void setConfigurationFile(String filename) {
   public synchronized void start() {
     if (scanDirectoriesFuture == null) {
       scanDirectoriesFuture =
-          scheduler.scheduleAtFixedRate(
+          scheduler.scheduleWithFixedDelay(
```
djcooke: Does this fix something? Doesn't it mean that Run Scanner would take even longer to get through all the runs after a restart?

apmasell: I'm hoping it will. There's a pattern in the Prometheus graphs that makes me think it's lapping itself, which isn't helpful. I'd like to avoid that. It shouldn't take any longer after start up. This will still process all the directories after the first scan.

PR #32, opened 2019-02-27
Beta PromethION Processor

avarsava:
Able to parse from FAST5:
- Run Alias
- Sequencer Folder Path
- Sequencer Position
- Container Serial Number
- Container Model
- Start Date
- Software version info

Gets Sequencer Name from JSON; Lane Count and Health Type are default values.
Traverses the deep directory structures of the PromethION storage for FAST5 files.


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultMinion.java
```diff
@@ -0,0 +1,29 @@
+package ca.on.oicr.gsi.runscanner.scanner.processor;
+
+import ca.on.oicr.gsi.runscanner.dto.NotificationDto;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.io.File;
+import java.io.IOException;
+import java.util.TimeZone;
+import java.util.stream.Stream;
+
+public class DefaultMinion extends RunProcessor {
```
apmasell: This should be held back until it has an implementation.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
```diff
@@ -87,7 +87,9 @@ public Platform getPlatformType() {
     Stream<Builder> standard =
         Stream.of(
             new Builder(Platform.ILLUMINA, "default", DefaultIllumina::create),
-            new Builder(Platform.PACBIO, "default", DefaultPacBio::create));
+            new Builder(Platform.PACBIO, "default", DefaultPacBio::create),
+            new Builder(Platform.PROMETHION, "default", DefaultPromethion::create),
```
apmasell: Rather than have two platforms that are `"default"`, make these both `OXFORDNANOPORE` and make one `"promethion"` and the other `"minion"`.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPromethion.java
```diff
@@ -0,0 +1,83 @@
...
+    } else { // root is a file
+      if (isFileFast5(root)) {
+        str.add(root);
+      }
+    }
+    return str.stream();
+  }
+
+  private boolean isFileFast5(File file) {
+    String fileName = file.getName();
```
apmasell: There's a `String.endsWith`.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/PromethionNotificationDto.java
```diff
@@ -0,0 +1,10 @@
+package ca.on.oicr.gsi.runscanner.dto;
+
+import ca.on.oicr.gsi.runscanner.dto.type.Platform;
+
```
apmasell: These two classes should be collapsed into one as well.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPacBio.java
```diff
@@ -353,4 +353,9 @@ private void processMetadata(Document metadata, PacBioNotificationDto dto) {
   private static boolean isStringBlankOrNull(String s) {
     return s == null || "".equals(s.trim());
   }
+
+  @Override
+  public boolean isFilePathValid(File fs_obj) {
+    return fs_obj.isDirectory() && fs_obj.canExecute() && fs_obj.canRead();
```
djcooke: Can put this in the abstract class, Javadoc the default behaviour, and only override if it should do something different (DefaultPromethion)

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPromethion.java
```diff
@@ -0,0 +1,81 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.time.ZonedDateTime;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.TimeZone;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class DefaultPromethion extends RunProcessor {
```
djcooke: Is this going to be PromethION-specific? Does there need to be a separate processor for MinION?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
```diff
@@ -190,4 +192,12 @@ public final Platform getPlatformType() {
    *     directory may be processed again.
    */
   public abstract NotificationDto process(File runDirectory, TimeZone tz) throws IOException;
+
+  /**
+   * Determine whether a File is readable by the processor.
+   *
+   * @param fs_obj File object which may represent a directory or file.
```
djcooke: Use camel-case and descriptive variable/parameter names. This will be a path that was specified in run scanner config, right?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
```diff
@@ -87,7 +87,9 @@ public Platform getPlatformType() {
     Stream<Builder> standard =
         Stream.of(
             new Builder(Platform.ILLUMINA, "default", DefaultIllumina::create),
-            new Builder(Platform.PACBIO, "default", DefaultPacBio::create));
+            new Builder(Platform.PACBIO, "default", DefaultPacBio::create),
+            new Builder(Platform.OXFORDNANOPORE, "promethion", DefaultPromethion::create) /*,
+            new Builder(Platform.OXFORDNANOPORE, "minion", DefaultMinion::create)*/);
```
djcooke: remove commented-out code

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/FindRuns.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * Attempts to process run directories, provided on the command line, through getRunsFromRoot() and
+ * display the results. This is for debugging purposes.
+ */
+public final class FindRuns {
+
+  public static void main(String[] args) throws IOException {
+    if (args.length == 0) {
+      System.err.println(
+          "Usage: java -DplatformType=ILLUMINA -Dname=default -Dtz=America/Toronto -Dparameters={} ca.on.oicr.gsi.runscanner.scanner.Test_getRunsFromRoot /path/to/run/folder");
```
djcooke: I think the class name in this usage message is wrong Please also document this utility in the readme

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPromethion.java
avarsava: As far as I can tell, the PromethION and MinION fast5 files use the same fields to store different information, and save them in different directory structures on disk. They will need separate RunProcessors.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
avarsava: This code will be uncommented when the MinION processor is ready.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPromethion.java
apmasell: It probably makes sense to have a `BaseNanoporeProcessor` that includes the method that handles the Fast5 and then have a `PromethionProcessor` that extends it. The `DefaultX` thing was because there was only one processor for Illumina and PacBio, but they may have had other instruments. Particularly, the Illumina GenomeAnalyzers would need a totally independent processor, but we decided not to bother writing one for that data.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPacBio.java
avarsava: When the Default MinION Processor is implemented, we will have a 50/50 split between reading directories vs files. Should directories still be considered default?

djcooke: Maybe it would be better to add a `PathType` enum `{DIRECTORY, FILE}` and `getPathType` abstract method in `RunProcessor`. Then the `isFilePathValid` method can be only in the abstract class, and check `getPathType` to determine which checks to do. This reduces duplication so that if, for example, you realize another permission that needs checked on run directories, you only need to make that change in one place.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
djcooke: We try to keep code clean of commented code and TODO notes as these make more sense in other branches and JIRA tickets respectively

README.md
```diff
@@ -97,15 +97,17 @@ If you would like to scan for Illumina output, please see [runscanner-illumina/R
...
+
+### Retrieving run output
 For troublesome runs, you can see the output for a particular run directory using:
 
     java -cp $RUN_SCANNER_HOME/WEB-INF/classes:$RUN_SCANNER_HOME/WEB-INF/lib/'*' ca.on.oicr.gsi.runscanner.scanner.Main
 
 It will display instructions on how to use it. You will have to set the `RUN_SCANNER_HOME` to the path containing an unpacked version of the WAR.
 
+### List observable runs
+To troubleshoot whether a processor can observe a run file, you can see a list of all the files a processor will accept from a directory using:
```
djcooke: Does it find all files in a directory, or all runs in a path?

avarsava: The files in the directory structure may not be of the right type (eg the PromethION processor only accepts .fast5 files, not .fastq). It lists all the files in the directory structure which the processor will process. (eg all the fast5 files, not listing any fastqs which may be present as well)

djcooke: For the PromethION processor, it will list all of the fast5 files within the specified directory (recursively), yes. For PromethION, a fast5 file is a run that it will process. For Illumina, it will list all of the subdirectories within the specified directory. For Illumina, the directory is the run. PacBio will be similar to Illumina. The only case for which it "lists all of the files that a processor will accept from a directory" is for PromethION, but in all cases, it will list all of the things that the processor will treat as runs.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPromethion.java
djcooke: > the PromethION and MinION fast5 files use the same fields to store different information, and save them in different directory structures on disk

:man_facepalming:

PR #33, opened 2019-03-01
Fix Testing.java

avarsava:


apmasell:
I think you want to rebase this...

avarsava:
> I think you want to rebase this...

I think I do. Will the rebase & merge button cover this situation when I get a second approval or do I have to do it earlier?

apmasell:
I would do it first yourself just to see what the commit history shakes out like.

PR #37, opened 2019-03-12
Beta MinION Processor

avarsava:
Parses from FAST5:
- Run Alias
- Container Serial Number
- Container Model
- Start time
- Software version

device_id doesn't seem to describe sequencerPosition for MinION, made that unique to PromethION, rest is shared.

### DIFF
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/BaseNanoporeProcessor.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/BaseNanoporeProcessor.java
@@ -126,9 +126,9 @@ public abstract class BaseNanoporeProcessor extends RunProcessor {
             + " + "
             + reader.getAttr(TRACKING_ID, "protocols_version"));
 
-    return processUnique(nnd, reader);
+    additionalProcess(nnd, reader);
+    return nnd;
   }
 
-  protected abstract NanoporeNotificationDto processUnique(
-      NanoporeNotificationDto nnd, IHDF5StringReader reader);
+  protected abstract void additionalProcess(NanoporeNotificationDto nnd, IHDF5StringReader reader);
 }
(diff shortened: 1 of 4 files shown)


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/MinionProcessor.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.stream.Stream;
+
+public class MinionProcessor extends BaseNanoporeProcessor {
+  public MinionProcessor(Builder builder, String seqName) {
+    super(builder, seqName);
+  }
+
+  @Override
+  protected Stream<Path> readsDirectoryForRun(Path path) {
+    return IntStream.of(0).mapToObj(i -> path.resolve(Paths.get("fast5", Integer.toString(i))));
```
apmasell: If this can only ever be one thing, just return `Stream.of(path.resolve(Paths.get("fast5", "0"))`

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/BaseNanoporeProcessor.java
```diff
@@ -29,4 +45,90 @@ public BaseNanoporeProcessor(Builder builder, String seqName) {
...
+        ZonedDateTime.parse(reader.getAttr(TRACKING_ID, "exp_start_time")).toLocalDateTime());
+    nnd.setSoftware(
+        reader.getAttr(TRACKING_ID, "version")
+            + " + "
+            + reader.getAttr(TRACKING_ID, "protocols_version"));
+
+    return processUnique(nnd, reader);
+  }
+
+  protected abstract NanoporeNotificationDto processUnique(
```
apmasell: I don't think this method needs to return since it's going to only mutate the object. If it returned a different instance of the DTO than the one it was given, that would be weird. Also, the name isn't great. You can call it just `process` if you like (since the overload resolution makes it unique). Maybe `additionalProcess` or something like that. `Unique` isn't conveying anything about the actions of this method.

PR #45, opened 2019-03-28
GLT-2777: added instructions for deploying to non-root context

djcooke:


djcooke:
@callunity: added note to refer to the description above. Also renamed "Releasing" section to "Deploying"

PR #48, opened 2019-04-03
[GLT-2789] Human-readable Run Aliases for ONT

avarsava:
Change runAlias from the hash in the fast5 to the name of the run directory (in the case of the PromethION, append the sequencer position) for the sake of readability. 
Also track the exp_script_purpose from the fast5 for the sake of distinguishing sequencing and mux runs on the MinION.

### DIFF
--- a/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/OxfordNanoporeNotificationDto.java
+++ b/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/OxfordNanoporeNotificationDto.java
@@ -3,18 +3,18 @@ package ca.on.oicr.gsi.runscanner.dto;
 import ca.on.oicr.gsi.runscanner.dto.type.Platform;
 
 public class OxfordNanoporeNotificationDto extends NotificationDto {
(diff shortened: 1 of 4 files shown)


dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/OxfordNanoporeNotificationDto.java
```diff
@@ -3,8 +3,18 @@
 import ca.on.oicr.gsi.runscanner.dto.type.Platform;
 
 public class OxfordNanoporeNotificationDto extends NotificationDto {
+  private String scriptPurpose;
```
callunity: What do you think about the term `runType` for this? (I don't know that `scriptPurpose` necessarily gives a huamn reading the DTO as much context for what the field means)

avarsava: Sounds good!

avarsava:
Run Aliases need another reconfiguration, won't merge until I do that

PR #50, opened 2019-04-23
[GLT-2789] Human-readable Run Names for PromethION

avarsava:
To test: point to local directory structure of:
date_name/nnnnn_runN/run_alias_here/fast5_fail
''/''/''/fast5_pass
''/''/''/fastq_fail
''/''/''/fastq_pass
''/''/''/sequencing_summary
one or more of which contains a valid fast5. Will use 'run_alias_here' as alias.

Also tracks script purpose to distinguish sequencing and mux runs


avarsava:
Performance issue. Not merging until I can figure it out

apmasell:
Looks good!

callunity:
Can any of the log messages be removed or are they all necessary?

avarsava:
> Can any of the log messages be removed or are they all necessary?

Non-error messages have been set to 'debug', which should not show at the default log4j output level.

dto/src/test/java/ca/on/oicr/gsi/runscanner/rs/dto/test/AbstractNotificationDtoTest.java
```diff
@@ -0,0 +1,54 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
+import com.fasterxml.jackson.databind.util.ISO8601DateFormat;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import java.time.LocalDateTime;
+import org.junit.Before;
+import org.junit.Test;
+
+public abstract class AbstractNotificationDtoTest {
+  NotificationDto notificationDto;
```
apmasell: Since doing this requires a lot of casting in the derived classes, can you make this a field in the children of the derived type and add an abstract method that returns `NotificationDto` that is used by these test?

scanner/src/test/resources/oxfordnanopore/promethion/20190221_2053_2-E3-H3_PAD10721_98c289ed/reference.json
```diff
@@ -0,0 +1 @@
+{"platform":"OxfordNanopore","runAlias":"20190221_2053_2-E3-H3_PAD10721_98c289ed","sequencerFolderPath":null,"sequencerName":"unittest","sequencerPosition":"2-E3-H3","containerSerialNumber":"PAD10721","containerModel":"flo-pro002","laneCount":1,"healthType":"UNKNOWN","startDate":[2019,2,21,20,54],"completionDate":null,"pairedEndRun":false,"software":null,"metrics":null,"runType":"sequencing_run"}
```
apmasell: We've been putting all of these throught `jq -S .` to format them and to minimise diffs when we update them in the future.

PR #57, opened 2019-05-24
Update Jackson Version

avarsava:
Updating just databind to 2.9.9 broke the tests (NoClassDef on
ObjectMapper) so now everything is at the same place.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,10 +1,3 @@
-# Unreleased
-
-Changes:
-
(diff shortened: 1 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -1,3 +1,10 @@
+# Unreleased
+
+Changes:
+
+ * Split MinKNOW and Protocol versions in ONT DTO
+ * Update Jackson versions all to 2.9.9 for security
```
djcooke: Not a user-facing change, so probably shouldn't be listed

PR #60, opened 2019-05-30
miso-lims#1483: fixed %>Q30 column labels; added density units

djcooke:
Based on Savo's input from https://github.com/miso-lims/miso-lims/issues/1483#issuecomment-397311508

### DIFF
--- a/runscanner-illumina/main.cpp
+++ b/runscanner-illumina/main.cpp
@@ -312,7 +312,7 @@ void add_lane_charts(
   add_table_column(columns, "Lane", "lane");
   add_table_column(columns, "Pool", "pool");
   add_table_column(columns, "Density %", "densityPct");
-  add_table_column(columns, "Density (K clusters/mm²)", "density");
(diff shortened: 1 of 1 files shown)


runscanner-illumina/main.cpp
```diff
@@ -312,9 +312,8 @@ void add_lane_charts(
   add_table_column(columns, "Lane", "lane");
   add_table_column(columns, "Pool", "pool");
   add_table_column(columns, "Density %", "densityPct");
-  add_table_column(columns, "Density", "density");
+  add_table_column(columns, "Density (K clusters/mm²)", "density");
```
apmasell: I have great fear that C++'s Unicode support will break, but here's hoping.

apmasell: Wait, there's a UTF-8 literal support: https://en.cppreference.com/w/cpp/language/string_literal
Try making this `u8"Density (K clusters/mm²)"`

PR #62, opened 2019-05-31
User Manual for Run Scanner

avarsava:
First version. Some parts aren't as thorough as I would like them but it is enough to release.

API page will be documented once API page is done (planning for separate PR for that)


docs/pages/appendices.md
```diff
@@ -0,0 +1,303 @@
...
+
+
+
+<a id="C" />
+### Appendix C: Sequencers Supported by Run Scanner
+This is a list of sequencers for which Run Scanner is capable of processing 
+output. 
+**Bold** items are verified to work with Run Scanner. 
+_Italic_ items are under development. Other items should work through having 
+equivalent output, however they have not been tested.
```
callunity: What do you think of rephrasing the `Other items [...]` sentence to something like `Other items are expected to work so long as the sequencer's output is in the same format as a sequencer whose name is in bold, but they have not been tested.` 
Either way, could it be put on a different line from the 'under development' line?

docs/pages/appendices.md
```diff
@@ -0,0 +1,303 @@
...
+* **NextSeq 500**
+* **NextSeq 550**
+* **NovaSeq 6000**
+* iSeq
+* MiniSeq
+
+#### Oxford Nanopore
+
+* **PromethION** - as of Release 19.01.1
+* _MinION - as of Release 18.12_
```
callunity: Could it be clarified what software the Release numbers refer to? Right now it could be read as being a run scanner release version!

docs/pages/illuminasetup.md
```diff
@@ -0,0 +1,84 @@
...
+    sudo apt-get install pkg-config cmake libjsoncpp-dev autoconf libtool build-essential clang-5.0
+
+On Fedora/Red Hat:
+
+    sudo yum groupinstall "Development Tools"
+    sudo yum install pkgconfig cmake jsoncpp-devel libtool autoconf clang
+
+Pull the Illumina Interop code via:
+
+    git submodule init && git submodule update
```
callunity: Do you need to `cd runscanner-illumina` before this line?

avarsava: Just tried with a fresh clone of runscanner and this command works appropriately from the project root. These and the installation instructions were mostly copied from the respective readme's

docs/pages/appendices.md
avarsava: Updated, please let me know what you think

docs/pages/appendices.md
avarsava: Updated

docs/pages/appendices.md
```diff
@@ -0,0 +1,298 @@
...
+**Bold** items are verified to work with Run Scanner. _Italic_ items are under development. Other items should work through having equivalent output, however they have not been tested. 
+If a sequencer is not included in this list, it should be assumed to be unsupported.
+
+#### PacBio
+* **RS**
+* **RS II**
+* _Sequel - Under development_
+
+#### Illumina
+* **HiSeq 1000**
```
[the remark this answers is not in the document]
avarsava: Should I add it to the 'might work but we haven't tested it' list, or just exclude it altogether?

avarsava:
Added a blank page on monitoring with Prometheus so @apmasell can write a guide on that

docs/pages/appendices.md
```diff
@@ -303,6 +303,6 @@ unsupported.
 
 #### Oxford Nanopore
 
-* **PromethION** - as of Release 19.01.1
-* _MinION - as of Release 18.12_
+* **PromethION** - as of PromethION Release 19.01.1
+* _MinION - as of MinION Release 18.12_
```
callunity: It looks like the "MinION" label is missing an underscore, but should it be double-asterisked instead to match the PromethION label?

docs/pages/appendices.md
callunity: Looks good to me!

docs/pages/appendices.md
avarsava: The MinION Processor isn't 100% done yet, it'll be bolded instead of italicized once I've verified it's done. I've italicized the whole line for PacBio Sequel and MinION so it's harder to miss

PR #66, opened 2019-06-06
Swagger documentation for REST API

avarsava:


dto/pom.xml
```diff
@@ -38,5 +38,9 @@
       <groupId>cisd</groupId>
       <artifactId>jhdf5</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-client</artifactId>
```
djcooke: Is this needed for something? I tried ripping it out of all 3 poms and the Swagger docs still work fine

pom.xml
```diff
@@ -33,6 +33,8 @@
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
     <jackson.version>2.9.9</jackson.version>
+    <spring.version>5.1.1.RELEASE</spring.version>
```
djcooke: Should use latest patch, which is `5.1.7.RELEASE`

pom.xml
```diff
@@ -173,6 +191,11 @@
       <id>ImageJ Public</id>
       <url>http://maven.imagej.net/content/repositories/public/</url>
     </repository>
+    <repository>
+      <id>jcenter-snapshots</id>
```
djcooke: Is this needed for something? The new dependencies should all be on Maven Central

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
```diff
@@ -41,10 +39,17 @@
...
-      value = "/run/{name}",
-      produces = {"application/json"})
-  public ResponseEntity<NotificationDto> getByName(@PathVariable("name") String id) {
+  @GetMapping("/run/{name}")
+  @ApiOperation(
+      value = "Get run by name",
+      response = ca.on.oicr.gsi.runscanner.dto.NotificationDto.class,
+      responseContainer = "ResponseEntity")
+  @ApiResponses({
+    @ApiResponse(code = 200, message = "dto"),
```
djcooke: `ApiResponse` messages should describe the reason for the status. 200 = "Success," 404 = "Run not found," etc.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
```diff
@@ -41,10 +39,17 @@
...
+    @ApiResponse(code = 404, message = "null")
+  }) // TODO: How do I use the HttpStatus naming instead of manually using the value?
```
djcooke: You'd need a constant int value, which means you can't use the HttpStatus or similar Status enums. You could create a class with the constants:

```
public class HttpStatusCodes {
  private static final int SUCCESS = 200;
  private static final int NOT_FOUND = 404;
  ...
}
```

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
```diff
@@ -102,11 +115,14 @@
...
-      value = "/runs/progressive",
-      produces = {"application/json"})
+  @PostMapping("/runs/progressive")
+  @ApiOperation(
+      value = "Progressive scan of runs",
+      response = ca.on.oicr.gsi.runscanner.dto.ProgressiveResponseDto.class)
   @ResponseBody
-  public ProgressiveResponseDto progressive(@RequestBody ProgressiveRequestDto request) {
+  public ProgressiveResponseDto progressive(
+      @ApiParam(value = "ProgressiveRequestDto object created by requesting client") @RequestBody
```
djcooke: "Object containing request options" might be a better description? All parameters are, by definition, created/provided by the client

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/SwaggerConfig.java
```diff
@@ -0,0 +1,55 @@
...
+  String projectVersion;
+
+  @Bean
+  public Docket api() {
+    return new Docket(DocumentationType.SWAGGER_2)
+        .select()
+        .apis(RequestHandlerSelectors.any())
+        .paths(PathSelectors.any())
+        .build()
+        .apiInfo(metaData());
```
djcooke: Swagger's default responses include unauthorized and forbidden codes, which can never actually happen with Run Scanner since there is no authorization. I think this is the line in Pinery's Swagger config that removes all of the default responses for `GET` requests: https://github.com/oicr-gsi/pinery/blob/master/pinery-ws/src/main/java/ca/on/oicr/pinery/ws/component/SwaggerConfig.java#L37

Should change the default responses for `POST` as well, as there won't be unauthorized/forbidden responses, and the regular response for `/runs/progressive` should actually be 200 (OK) rather than 201 (Created)

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
```diff
@@ -57,8 +62,14 @@
    * Given a known run name. get its metrics. If no run is found, null is returned. If there are
    * multiple runs with the same name that are from different sequencers, one is randomly selected.
    */
-  @GetMapping(value = "/run/{name}/metrics")
-  public HttpEntity<byte[]> getMetricsByName(@PathVariable("name") String id) {
+  @GetMapping("/run/{name}/metrics")
+  @ApiOperation(
+      value = "Get metrics by run name",
+      response = byte[].class,
```
apmasell: The response will be a JSON array (even though it's handled as a byte array).

pom.xml
```diff
@@ -111,41 +122,47 @@
...
+        <version>2.9.2</version>
+      </dependency>
+      <dependency>
+        <groupId>io.springfox</groupId>
+        <artifactId>springfox-swagger-ui</artifactId>
+        <version>2.9.2</version>
+      </dependency>
+      <dependency>
+        <groupId>org.jboss.resteasy</groupId>
+        <artifactId>resteasy-client</artifactId>
```
djcooke: Shouldn't need RESTEasy here either, or in `scanner/pom.xml`

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
```diff
@@ -57,8 +63,14 @@
...
    * multiple runs with the same name that are from different sequencers, one is randomly selected.
    */
-  @GetMapping(value = "/run/{name}/metrics")
-  public HttpEntity<byte[]> getMetricsByName(@PathVariable("name") String id) {
+  @GetMapping("/run/{name}/metrics")
+  @ApiOperation(
+      value = "Get metrics by run name",
+      response = Json[].class,
+      responseContainer = "HttpEntity")
+  @ApiResponses({@ApiResponse(code = 200, message = "Success")})
```
djcooke: This endpoint responds with an empty array if the run doesn't exist, so the docs are correct in showing success only. It should be changed to return 404 status if the run isn't found though.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
```diff
@@ -68,19 +68,23 @@
...
-                .getBytes(StandardCharsets.UTF_8));
+    byte[] response =
+        scheduler
+            .finished()
+            .filter(dto -> dto.getRunAlias().equals(id))
+            .findAny()
+            .map(NotificationDto::getMetrics)
+            .orElse("[]")
+            .getBytes(StandardCharsets.UTF_8);
+    return (response.length > 0
```
djcooke: I think `"[]".getBytes(StandardCharsets.UTF_8).length will be > 0` too, no? This is checking the length of the byte array rather than the JSON array.

You can probably return `...findAny().map(...).getBytes(...).orElseGet(() -> new ResponseEntity<>(HttpStatus.NOT_FOUND))`, similar to the `/run/{name}` endpoint

PR #70, opened 2019-06-14
New Run Alias scheme for ONT

avarsava:
Now uses all directory names between root and fast5 location. Track root
location in OxfordNanopore RunProcessor to accomodate this. Alter ONT
Sequencer tests because they never call getRunsFromRoot().

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,7 +3,8 @@
 Changes:
 
   * Added support for MinION output
-  * Add runscanner-mystery-files.log 
+  * Add runscanner-mystery-files.log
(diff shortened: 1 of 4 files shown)


avarsava:
Dillan pointed out that this breaks the links to Run JSON. Will figure out new naming scheme with Lars and push an update, no need to review until that's done

apmasell:
Eh, substitute ` ` or `:` for `/` and call it a day.

avarsava:
@apmasell that's likely what we'll go with, just need Lars' OK first

avarsava:
Lars has requested underscores to match the reports

PR #75, opened 2019-07-04
Handle time as Instant, report datetime strings rather than arrays

avarsava:


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
```diff
@@ -221,7 +228,39 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+                return ZonedDateTime.parse(inststr).toInstant();
+              } catch (DateTimeParseException dtpe) {
+                throw new IOException(dtpe);
+              }
+            }
+          });
+
+      dto =
+          new ObjectMapper()
+              .registerModule(module)
```
djcooke: Did you look at the `JavaTimeModule` that might work instead? https://github.com/FasterXML/jackson-modules-java8

If not, could this module be defined in its own class rather than recreating it for every run?

It would also be ideal to have a single `ObjectMapper` configured via Spring and used throughout the application, so it behaves the same everywhere.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
```diff
@@ -298,16 +337,14 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
                     }
                     Matcher m = FAILED_MESSAGE.matcher(failMessage);
                     // Somehow, scanner will return things that don't match, so, we check again
-                    return m.matches()
-                        ? LocalDateTime.parse(m.group(1), FAILED_MESSAGE_DATE_FORMATTER)
-                        : null;
+                    return m.matches() ? Instant.parse(m.group(1)) : null;
```
apmasell: This is not going to be in the format that is match. You're still going to have to use the `FAILED_MESSAGE_DATE_FORMATTER` to get a `LocalDateTime` and then convert it.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPacBio.java
```diff
@@ -246,6 +256,7 @@ protected StatusResponse getStatus(String url) {
 
   @Override
   public NotificationDto process(File runDirectory, TimeZone tz) throws IOException {
+    this.timeZone = tz;
```
apmasell: Don't put this is in a field. We've already had concurrency issues this with. Pass it as a parameter to the relevant methods.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPacBio.java
```diff
@@ -300,17 +311,23 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
       builder.addParameter("instrument", dto.getSequencerName());
       builder.addParameter("run", dto.getRunAlias());
       builder.addParameter(
-          "from", dto.getStartDate().truncatedTo(ChronoUnit.DAYS).format(URL_DATE_FORMAT));
+          "from",
+          dto.getStartDate()
+              .truncatedTo(ChronoUnit.DAYS)
```
apmasell: I think the order needs to swap: put it in the right zone, then truncate.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
avarsava: All the other RunProcessors register an ObjectMapper with the JavaTimeModule. However with the Instant parsing that comes it, it requires that datetime strings be in UTC time, which is not possible to get from the C++ runscanner-illumina output. This custom de/serializer allows for the use of any timezone.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
```diff
@@ -221,7 +228,39 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+          new JsonDeserializer<Instant>() {
+            @Override
+            public Instant deserialize(
+                JsonParser jsonParser, DeserializationContext deserializationContext)
+                throws IOException, JsonProcessingException {
+              String inststr = jsonParser.getText();
+              try {
+                return ZonedDateTime.parse(inststr).toInstant();
+              } catch (DateTimeParseException dtpe) {
+                throw new IOException(dtpe);
```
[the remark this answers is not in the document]
avarsava: "'JsonProcessingException(java.lang.Throwable)' has protected access in 'com.fasterxml.jackson.core.JsonProcessingException'"

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
avarsava: Done

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultPacBio.java
avarsava: Done

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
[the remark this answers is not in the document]
djcooke: Use `JsonParseException` (extends `JsonProcessingException`)

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
djcooke: Ah, makes sense. Maybe just separate creating the module into a separate method as `process` is already really long

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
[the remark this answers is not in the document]
avarsava: Done

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
avarsava: Done

PR #79, opened 2019-08-02
Hiseq workflow type

apmasell:


pom.xml
```diff
@@ -32,7 +32,7 @@
     <skipIllumina>true</skipIllumina>
     <maven.compiler.source>1.8</maven.compiler.source>
     <maven.compiler.target>1.8</maven.compiler.target>
-    <jackson.version>2.9.9</jackson.version>
+    <jackson.version>2.9.9.2</jackson.version>
```
callunity: Will we be affected by this? https://github.com/FasterXML/jackson-databind/issues/2395

apmasell: Probably. I will rollback.

callunity:
Failing test-- expected workflowType null, got workflowType None

apmasell:
Yeah. Will look at it.

PR #87, opened 2019-10-01
Try new attribute name, try again with old name

avarsava:
Fixes GLT-3002. 'Unreadable' & stack trace when neither attribute is found maintains old behaviour

### DIFF
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/PromethionProcessor.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/PromethionProcessor.java
@@ -5,7 +5,6 @@ import ch.systemsx.cisd.hdf5.IHDF5Reader;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import java.nio.file.Path;
 import java.util.stream.Stream;
-import ncsa.hdf.hdf5lib.exceptions.HDF5AttributeException;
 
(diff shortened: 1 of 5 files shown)


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/PromethionProcessor.java
```diff
@@ -32,7 +33,12 @@ protected boolean excludedDirectoryFormat(Path path) {
   @Override
   protected void additionalProcess(OxfordNanoporeNotificationDto onnd, IHDF5Reader reader) {
     onnd.setSequencerPosition(reader.string().getAttr(trackingId, "device_id"));
-    onnd.setContainerModel(reader.string().getAttr(contextTags, "flowcell_type"));
+    try {
+      onnd.setContainerModel(reader.string().getAttr(contextTags, "flow_cell_product_code"));
+    } catch (HDF5AttributeException h5ae) {
+      onnd.setContainerModel(reader.string().getAttr(contextTags, "flowcell_type"));
+      // Will throw again, causing failed read & trace in log, if neither works
+    }
```
djcooke: If the exception is thrown because the attribute doesn't exist, can you use `reader#hasAttribute` to check for it before the first `#getAttr` instead of catching the exception? I'm not sure if there are other reasons the exception might be thrown, but better to avoid using exceptions to control regular behaviour when possible.

avarsava:
@djcooke release notes updated, will fix commit message when squashing

PR #88, opened 2019-10-22
GLT-3022: get ONT sequencer name from files

djcooke:


.settings/org.eclipse.m2e.core.prefs
```diff
@@ -0,0 +1,4 @@
+activeProfiles=
```
avarsava: Committing editor preference files, maybe update .gitignore?

djcooke: Nope, that's intentional. This way, everyone gets consistent formatting, etc. Unfortunately the formatter isn't configured properly for this project, but that's a separate issue

PR #90, opened 2019-10-29
GLT-3043: correct MinION sequencing container models

djcooke:


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/MinionProcessor.java
```diff
@@ -48,8 +48,10 @@ protected void additionalProcess(OxfordNanoporeNotificationDto onnd, IHDF5Reader
     if (reader.hasAttribute(trackingId, "device_id")) {
       onnd.setSequencerName(reader.string().getAttr(trackingId, "device_id"));
     }
-    if (reader.hasAttribute(contextTags, "sequencing_kit")) {
-      onnd.setContainerModel(reader.string().getAttr(contextTags, "sequencing_kit"));
+    if (reader.hasAttribute(contextTags, "flow_cell_product_code")) {
```
avarsava: PromethionProcessor has the exact same code, could this be moved to the process() in BaseOxfordNanoporeProcessor?

djcooke: Moved

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/MinionProcessor.java
```diff
@@ -48,9 +48,6 @@ protected void additionalProcess(OxfordNanoporeNotificationDto onnd, IHDF5Reader
     if (reader.hasAttribute(trackingId, "device_id")) {
       onnd.setSequencerName(reader.string().getAttr(trackingId, "device_id"));
     }
-    if (reader.hasAttribute(contextTags, "sequencing_kit")) {
-      onnd.setContainerModel(reader.string().getAttr(contextTags, "sequencing_kit"));
```
avarsava: Will getting rid of the check for sequencing_kit as a last resort break older fast5s, or was it never correct in the first place?

djcooke: It was never correct. `sequencing_kit` is the sequencing kit, not the flow cell model. See JIRA ticket for more details

PR #99, opened 2020-02-12
Try to use final_summary.txt for ONT runs

apmasell:
So, I don't know what the file looks like while the sequencer is running. I'm making some guesses and I'll try to observe the instrument running and get a better idea.


apmasell:
I'm not sure what happens if the run is paused or restarted.

PR #103, opened 2020-04-06
Don't accept "" strings

avarsava:


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/BaseOxfordNanoporeProcessor.java
```diff
@@ -222,34 +222,46 @@ public synchronized NotificationDto process(File runDirectory, TimeZone tz) thro
       trackingId = read_name + "/tracking_id";
       contextTags = read_name + "/context_tags";
 
-      if (genericReader.hasAttribute(trackingId, "flow_cell_id"))
-        onnd.setContainerSerialNumber(genericReader.string().getAttr(trackingId, "flow_cell_id"));
+      if (genericReader.hasAttribute(trackingId, "flow_cell_id")) {
+        String id = genericReader.string().getAttr(trackingId, "flow_cell_id");
+        if (!(id.equals(""))) onnd.setContainerSerialNumber(id);
+      }
```
djcooke, line 228: Could refactor this into a method since it's repeated several times:

```
protected void getAttribute(IHDF5Reader genericReader, String attribute, Consumer<String> setter)
```

PR #114, opened 2021-01-18
No more sequel

avarsava:
and also clarify the README bc it threw me off

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -14,7 +14,7 @@ Install [mkdocs](https://www.mkdocs.org/#installation).
 
 ## Test and deploy locally
 
-To test out changes locally, run:
+To test out changes locally, run from /runscanner:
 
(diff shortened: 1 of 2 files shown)


apmasell (approved):

docs/README.md
```diff
@@ -14,7 +14,7 @@ Install [mkdocs](https://www.mkdocs.org/#installation).
 
 ## Test and deploy locally
 
-To test out changes locally, run:
+To test out changes locally, run from /runscanner:
```
djcooke, line 17: "from the repository root directory" might be more accurate as it doesn't have to be named `runscanner`

djcooke (approved):

docs/README.md
avarsava: Fixed

PR #156, opened 2024-04-12
GLT-3648 exclude runs from runscanner

Christopher-Ho:
JIRA Ticket: https://jira.oicr.on.ca/browse/GLT-3648

- [ X ] Updates release notes
- Bad runs gauge metric is already in use for unreadable runs (alerting)
- Added Good runs gauge metric for readable runs (not for alerting)
- Added ability to exclude runs by adding directory name to ignore subdirectory list in configuration file. 
- Add ignored subdirectories entries on a newline to avoid horizontal scrolling due to amount of directories 
- Handle cases where configuration doesn't include ignore subdirectories field
- Added unit tests for skipSubDirectory

### DIFF
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
@@ -98,7 +98,7 @@ public class Scheduler {
     private String path;
     private Platform platformType;
     private String timeZone;
-    private ArrayList<File> ignoreSubdirectories;
+    private List<File> ignoreSubdirectories;
 
     public String getName() {
       return name;
@@ -120,7 +120,7 @@ public class Scheduler {
       return timeZone;
     }
 
-    public ArrayList<File> getIgnoreSubdirectories() {
+    public List<File> getIgnoreSubdirectories() {
       return ignoreSubdirectories;
     }
 
@@ -382,12 +382,31 @@ public class Scheduler {
     return scanningNow;
   }
 
-  // Helper to check if current run directory should be ignored
-  private boolean skipSubDirectory(
-      File currentRunDirectory, ArrayList<File> ignoreDirectories, File baseDirectory) {
-    for (File currentSubDirectory : ignoreDirectories) {
-      File basePlusSubDirectory = new File(baseDirectory, currentSubDirectory.getPath());
+  private static boolean isSubDirectory(File baseDirectory, File subDirectory) {
+    File parent = subDirectory.getParentFile();
+    while (parent != null) {
+      if (baseDirectory.equals(parent)) {
+        return true;
+      }
+      parent = parent.getParentFile();
+    }
+    return false;
+  }
+
+  /** Helper to check if current run directory should be ignored */
+  protected static boolean skipSubDirectory(
+      File currentRunDirectory, List<File> ignoreDirectories, File baseDirectory) {
+    // Loop through list of directories to ignore
+    for (File currentIgnoreDirectory : ignoreDirectories) {
+      File basePlusSubDirectory = new File(baseDirectory, currentIgnoreDirectory.getPath());
+
+      // Check if current directory is same as ignore directory
       if (currentRunDirectory.equals(basePlusSubDirectory)) {
+        return true;
+      }
+
+      // Check if current directory is subdirectory of ignore directory
+      if (isSubDirectory(basePlusSubDirectory, currentRunDirectory)) {
         return true;
       }
     }
@@ -401,14 +420,14 @@ public class Scheduler {
    * list of subdirectories to ignore nor needs reprocessing (for runs still active on the
    * sequencer)
    */
-  private boolean isUnprocessed(
-      File directory, ArrayList<File> ignoreDirectories, File baseDirectory) {
+  private boolean isUnprocessed(File directory, List<File> ignoreDirectories, File baseDirectory) {
     return !workToDo.contains(directory)
         && !processing.contains(directory)
         && (!failed.containsKey(directory)
             || Duration.between(failed.get(directory), Instant.now()).toMinutes() > 20)
         && (!finishedWork.containsKey(directory) || finishedWork.get(directory).shouldRerun())
-        // Exclude from processing if directory name in list of directories to ignore directory
+        // Exclude from processing if directory name in list of directories to ignore,
+        // or it is a sub-directory of an ignore directory
         && !skipSubDirectory(directory, ignoreDirectories, baseDirectory);
   }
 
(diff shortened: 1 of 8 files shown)


avarsava:
This example from the jira ticket, is this still current to how the config should look? 
image

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
```diff
@@ -137,6 +143,10 @@ public void setPlatformType(Platform platformType) {
     public void setTimeZone(String timeZone) {
       this.timeZone = timeZone;
     }
+
+    public void setIgnoreSubdirectories(ArrayList<String> ignoreSubdirectories) {
+      this.ignoreSubdirectories = ignoreSubdirectories; // Addition !!!
```
avarsava, line 148: What do these two 'addition' comments mean?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
```diff
@@ -361,15 +385,18 @@ public boolean isScanningNow() {
...
    */
   private boolean isUnprocessed(File directory) {
     return !workToDo.contains(directory)
         && !processing.contains(directory)
         && (!failed.containsKey(directory)
             || Duration.between(failed.get(directory), Instant.now()).toMinutes() > 20)
-        && (!finishedWork.containsKey(directory) || finishedWork.get(directory).shouldRerun());
+        && (!finishedWork.containsKey(directory) || finishedWork.get(directory).shouldRerun())
+        // Exclude from processing if directory name in list of directories to ignore
+        && roots.stream().noneMatch(n -> n.getIgnoreSubdirectories().contains(directory.getName()));
```
avarsava, line 399: Can't build runscanner to test right now - isUnprocessed is called as a filter while streaming all the roots (which is a List\<Configuration\>), this method also streams all the roots, and it doesn't look like ignoreSubdirectories is storing the absolute path. Is there interference across Configurations? If, for example, 'test' is listed as an ignoredSubdirectory for a Miseq, would it also get ignored for a NovaSeq (held in a different configuration)?

avarsava:
> > This example from the jira ticket, is this still current to how the config should look? image
> 
> Yes that example image is how the configuration looks like in the current implementation.

So if I had the directory structure:
```
/home/avarsava/nextseq/abcd/RECYCLE
/home/avarsava/nextseq/efgh/RECYCLE
/home/avarsava/nextseq/ijkl/RECYCLE
etc.
``` 
would adding 'RECYCLE' to 'ignoreSubdirectories' be sufficient? what if i wanted to scan `/home/avarsava/nextseq/RECYCLE` or even `/home/avarsava/nextseq/zyxw/RECYCLE` but not any of the rest?

djcooke:
> > > This example from the jira ticket, is this still current to how the config should look? image
> > 
> > 
> > Yes that example image is how the configuration looks like in the current implementation.
> 
> So if I had the directory structure:
> 
> ```
> /home/avarsava/nextseq/abcd/RECYCLE
> /home/avarsava/nextseq/efgh/RECYCLE
> /home/avarsava/nextseq/ijkl/RECYCLE
> etc.
> ```
> 
> would adding 'RECYCLE' to 'ignoreSubdirectories' be sufficient? what if i wanted to scan `/home/avarsava/nextseq/RECYCLE` or even `/home/avarsava/nextseq/zyxw/RECYCLE` but not any of the rest?

I don't think that's how it should work..? The path specified to ignore should be relative to the configured directory.

If you had a base directory `/home/avarsava/nextseq` and wanted to ignore `/home/avarsava/nextseq/abcd`, then that should be specified as `abcd`, which would work with the current implementation.

If you wanted to ignore `/home/avarsava/nextseq/abcd/RECYCLE`, I'd think that should be configured as `abcd/RECYCLE` so as not to have unintended effects... That won't work with the current implementation, but is there is a need to go ignore more than 1 level deep into the base directory? If there's some garbage directory inside a run directory, that's fine. This is just to prevent it from detecting something as a run directory.

avarsava:
> is there is a need to go ignore more than 1 level deep into the base directory?

Potentially? I'm unclear on what problem this feature is meant to solve, but other users might have needs different from ours.
My question was more 'does adding RECYCLE, as in the example config, count as \*/RECYCLE/\* and if not should it'

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
avarsava: right, so for the following configuration file:
```
[
  {
    "name": "nextseq",
    "parameters": {},
    "path": "/home/avarsava/nextseq/",
    "platformType": "ILLUMINA",
    "timeZone": "America/Toronto",
    "ignoreSubdirectories": [
      "RECYCLE"
    ]
  },
  {
    "name": "miseq",
    "parameters": {},
    "path": "/home/avarsava/miseq/",
    "platformType": "ILLUMINA",
    "timeZone": "America/Toronto",
    "ignoreSubdirectories": []
  }
]
```
This will create 2 Configuration objects which are streamed by 'roots.stream()'. `/home/avarsava/miseq/RECYCLE' would match by this logic even though it's not specified for exclusion, would it not?

djcooke: Yep. It's currently checking the `ignoreSubdirectories` for every root instead of only the specific root that it belongs to

djcooke:
> > is there is a need to go ignore more than 1 level deep into the base directory?
> 
> Potentially? I'm unclear on what problem this feature is meant to solve, but other users might have needs different from ours. My question was more 'does adding RECYCLE, as in the example config, count as */RECYCLE/* and if not should it'

This is basically so we can ignore things that are currently in unreadable: https://runscanner-stage.gsi.oicr.on.ca/listUnreadable

Note that each is a direct subdirectory of one of the configured root directories. Anything deeper than that is considered part of the run data. A run processor should be looking at specific files within a run directory, so there shouldn't be a need to explicitly ignore anything at that level

djcooke:
> Are we okay with the naming `ignoreSubdirectories` or should I consider something else like `ignoreRunDirectories`?

Some of the directories we'll want to ignore are not actually run directories, so I think just calling them subdirectories is more accurate

djcooke:
Looks like there is a need to allow ignoring deeper subdirectories - the Oxford Nanopore processors support nested run directories. This means you'll have to check the full path rather than just `directory.getName()` since you'd want to be able to ignore `$ROOT/run1/run1` without ignoring `$ROOT/run1`

avarsava (approved):

PR #164, opened 2024-09-10
[GP-4411] Endpoint to remove the cache for a run by name

avarsava:
JIRA Ticket: GP-4411

- [X] Updates release notes

### DIFF
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
@@ -66,16 +66,16 @@ public class RestResponseController {
   @DeleteMapping("/run/{name}")
   @ApiOperation(
       value = "Invalidate cache for run by name",
-      response = Boolean.class,
+      response = Void.class,
       responseContainer = "ResponseEntity")
   @ApiResponses({
-    @ApiResponse(code = 200, message = "Success"),
+    @ApiResponse(code = 204, message = "Success"),
     @ApiResponse(code = 404, message = "Run not found")
   })
-  public ResponseEntity<Boolean> deleteByName(
+  public ResponseEntity<Void> deleteByName(
       @PathVariable("name") @ApiParam(value = "Run name") String id) {
     if (scheduler.invalidate(id)) {
-      return new ResponseEntity<>(HttpStatus.OK);
+      return new ResponseEntity<>(HttpStatus.NO_CONTENT);
     } else {
       return new ResponseEntity<>(HttpStatus.NOT_FOUND);
     }
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
@@ -64,21 +64,14 @@ public class RestResponseController {
   }
 
   @DeleteMapping("/run/{name}")
-  @ApiOperation(
-      value = "Invalidate cache for run by name",
-      response = Void.class,
-      responseContainer = "ResponseEntity")
+  @ApiOperation(value = "Invalidate cache for run by name")
   @ApiResponses({
     @ApiResponse(code = 204, message = "Success"),
     @ApiResponse(code = 404, message = "Run not found")
   })
-  public ResponseEntity<Void> deleteByName(
-      @PathVariable("name") @ApiParam(value = "Run name") String id) {
-    if (scheduler.invalidate(id)) {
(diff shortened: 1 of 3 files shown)


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
```diff
@@ -63,6 +63,24 @@ public ResponseEntity<NotificationDto> getByName(
...
+  @DeleteMapping("/run/{name}")
+  @ApiOperation(
+      value = "Invalidate cache for run by name",
+      response = Boolean.class,
+      responseContainer = "ResponseEntity")
+  @ApiResponses({
+    @ApiResponse(code = 200, message = "Success"),
+    @ApiResponse(code = 404, message = "Run not found")
+  })
+  public ResponseEntity<Boolean> deleteByName(
```
djcooke, line 75: This isn't actually returning a boolean value, but I don't think it should anyway since the response status already indicates the result. Should return 204 (no content) for success though

djcooke (commented):
API docs look fine to me. Was there something that looked wrong to you?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
avarsava: Swagger does say "Resolver error at paths./runscanner/run/{name}.delete.responses.204.schema.$ref
Could not resolve reference: Could not resolve pointer: /definitions/Error-ModelName{namespace='java.lang', name='Void'} does not exist in document" when I do this though

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
```diff
@@ -382,6 +382,16 @@ public boolean isScanningNow() {
...
   }
 
+  public boolean invalidate(String runName) {
+    int oldSize = finishedWork.size();
+    finishedWork
+        .keySet()
+        .stream()
+        .filter(file -> file.getName().equals(runName))
+        .forEach(finishedWork::remove);
+    return finishedWork.size() < oldSize;
```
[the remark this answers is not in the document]
avarsava: Updated, not sure if this is what you meant though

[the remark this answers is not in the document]
djcooke: No, this part is the problem:

```
return finishedWork.size() < oldSize;
```

It's possible that other runs will be added to `finishedWork` while you are removing the intended run, so `finishedWork.size()` may be >= `oldSize` even after successfully removing the run.

My suggestion was to return the end of the stream with `.anyMatch(Objects::nonNull)` as that would indicate an item was removed as intended. I don't think looping on a failed remove is helpful as that shouldn't even be possible mid-stream after filtering.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
```diff
@@ -63,6 +63,24 @@ public ResponseEntity<NotificationDto> getByName(
...
+    @ApiResponse(code = 404, message = "Run not found")
+  })
+  public ResponseEntity<Void> deleteByName(
+      @PathVariable("name") @ApiParam(value = "Run name") String id) {
+    if (scheduler.invalidate(id)) {
+      return new ResponseEntity<>(HttpStatus.NO_CONTENT);
+    } else {
+      return new ResponseEntity<>(HttpStatus.NOT_FOUND);
+    }
+  }
```
djcooke, line 82: You can make this simpler by removing the `ResponseEntity` and just making the method return `void`. Throw `ResourceNotFoundException` if the run isn't found. I'm guessing for Swagger, you'd just remove the `response` and `responseContainer` parameters since it's not returning anything

djcooke: You'd have to add Spring's `@ResponseStatus` annotation to the method to make it return 204 with this change

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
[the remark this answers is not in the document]
avarsava: Got it, thanks!

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/RestResponseController.java
avarsava: Fixed, thanks!

djcooke (approved):

PR #165, opened 2024-09-13
[GP-4334] Parse in DRAGEN's BCLConvert into run information

avarsava:
JIRA Ticket: GP-4334

- Introduces NovaSeqXProcessor (open to changing the name) which parses (currently just) BCLConvert information from the Analysis directory and provides it in a new 'analysis' json block
- Shifts responsibility of calculating 'doneness' to the individual DTOs, with healthType remaining the standard, so that IlluminaNotificationDto can also take the new analysisStatus into account (as analysis may finish after sequencing is all done)
- Some outstanding questions in TODO
- Probably need to add tests

- [ ] Updates release notes


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
```diff
@@ -0,0 +1,241 @@
...
+              })
+          .forEach(a -> ret.add(a.toJson()));
+      return ret;
+    }
+  }
+
+  final String NUMERAL = "\\d+";
+  static final ObjectMapper MAPPER = new ObjectMapper();
+
+  BCLConvertAnalysis BCLConvertAnalyses = new BCLConvertAnalysis();
```
djcooke, line 91: This doesn't belong to the processor class - it should be kept within the `process` method

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
```diff
@@ -0,0 +1,241 @@
...
+                  return Integer.compare(a1.lane, a2.lane);
+                }
+                return comp;
+              })
+          .forEach(a -> ret.add(a.toJson()));
+      return ret;
+    }
+  }
+
+  final String NUMERAL = "\\d+";
```
djcooke, line 88: Should be private. Several other fields and methods within the PR missing access modifiers where package private (default) doesn't make the most sense

djcooke (commented):
Is there a reason this needs to be a separate processor? Not all NovaSeq X will have DRAGEN enabled, and it's not specific to NovaSeq X either. NextSeqs already have DRAGEN, and I'm sure Illumina will be adding it to other platforms as well. It probably makes sense to make it available via `DefaultIllumina`.

Is there any way to determine whether any analysis is expected for a run? If not, maybe there should just be a setting in the config to enable/disable any possible workflows?

avarsava:
> Is there a reason this needs to be a separate processor? Not all NovaSeq X will have DRAGEN enabled, and it's not specific to NovaSeq X either. NextSeqs already have DRAGEN, and I'm sure Illumina will be adding it to other platforms as well. It probably makes sense to make it available via DefaultIllumina.

I wanted to separate it out so as to not slow down scanning the older instruments. NextSeqs do already have DRAGEN, however, I don't trust Illumina to output the same way on different sequencers, and testing against the NextSeq is not in scope for this project. @mlaszloffy I'm pretty sure the goal is for all NovaSeq X runs to include DRAGEN? Can you confirm?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
```diff
@@ -0,0 +1,241 @@
...
+              switch (sectionName) {
+                case "[BCLConvert_Data]":
+                  if (line[0].equals("Lane")) break; // Skip column label line
+                  s++; // Hopefully the lines stream in a consistent order and this will recreate
+                  // the S#
+                  ObjectNode nested = MAPPER.createObjectNode();
+                  nested.put("Lane", line[0]);
+                  nested.put("Sample_ID", line[1]);
+                  nested.put("Index", line[2]);
+                  nested.put("Index2", line[3]); // Some have a line[4] like "Y27;I10;I10;Y27"
```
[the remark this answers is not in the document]
avarsava: For exporting the sample sheet (which this code is doing), I think it will be more flexible going forward to keep the same formatting as the sample sheet rather than needing to convert between two sets of keys

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/type/DRAGENWorkflow.java
```diff
@@ -0,0 +1,5 @@
+package ca.on.oicr.gsi.runscanner.dto.type;
+
+public enum DRAGENWorkflow {
+  BCLConvert
+}
```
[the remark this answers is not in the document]
avarsava: Illumina has been pretty consistent about the product being DRAGEN, not Dragen

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
```diff
@@ -0,0 +1,241 @@
...
+    Analysis get(String sample, String lane) {
+      return get(sample, Integer.parseInt(lane));
+    }
+
+    Analysis get(String sample, int lane) {
+      return analyses
+          .stream()
+          .filter(a -> a.sample.equals(sample) && a.lane == lane)
+          .findFirst()
+          .orElse(new Analysis());
```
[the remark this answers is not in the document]
avarsava: This is the only way new Analysis objects are created. All attempts to write to the analysis cache start with calling this get(), which gets either a previously-written Analysis or a new one if that doesn't exist, and then the updated Analysis is put() back into the cache.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
```diff
@@ -0,0 +1,241 @@
...
+import java.nio.file.Files;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.TimeZone;
+
+public class NovaseqXProcessor extends DefaultIllumina {
+  static class BCLConvertAnalysis {
```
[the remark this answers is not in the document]
avarsava: I had kept everything to one file to avoid an explosion of DRAGEN-related files in the processor package. I can change this if that's preferred though

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
```diff
@@ -0,0 +1,241 @@
...
+          .filter(a -> a.sample.equals(sample) && a.lane == lane)
+          .findFirst()
+          .orElse(new Analysis());
+    }
+
+    void put(Analysis newAnalysis) {
+      Analysis oldAnalysis = get(newAnalysis.sample, String.valueOf(newAnalysis.lane));
+      if (!oldAnalysis.isEmpty()) {
+        analyses.remove(oldAnalysis);
+      }
```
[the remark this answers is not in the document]
avarsava: The scan of the fastq_list file creates the first version of the Analysis object, then if there is a Demultiplex_Stats file, the updated Analysis objects created there need to overwrite the old ones

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
```diff
@@ -0,0 +1,241 @@
...
+    Analysis get(String sample, int lane) {
+      return analyses
+          .stream()
+          .filter(a -> a.sample.equals(sample) && a.lane == lane)
+          .findFirst()
+          .orElse(new Analysis());
+    }
+
+    void put(Analysis newAnalysis) {
+      Analysis oldAnalysis = get(newAnalysis.sample, String.valueOf(newAnalysis.lane));
```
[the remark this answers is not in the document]
avarsava: Because I added that signature for that method to address this line in particular then forget to actually switch to using it. :sweat: Fixed!

mlaszloffy:
> > Is there a reason this needs to be a separate processor? Not all NovaSeq X will have DRAGEN enabled, and it's not specific to NovaSeq X either. NextSeqs already have DRAGEN, and I'm sure Illumina will be adding it to other platforms as well. It probably makes sense to make it available via DefaultIllumina.
> 
> I wanted to separate it out so as to not slow down scanning the older instruments. NextSeqs do already have DRAGEN, however, I don't trust Illumina to output the same way on different sequencers, and testing against the NextSeq is not in scope for this project. @mlaszloffy I'm pretty sure the goal is for all NovaSeq X runs to include DRAGEN? Can you confirm?

Most future runs will have run with DRAGEN, but not necessarily all. For example, we may sequence for another team or facility and not have the sample information to perform DRAGEN analysis.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/type/DRAGENWorkflow.java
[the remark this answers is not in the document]
djcooke: Documentation should use Illumina's convention. Java code should use Java convention

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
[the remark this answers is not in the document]
djcooke: It's not going to be a lot of files, though it doesn't really matter if it is. You can make a sub-package to keep all the analysis/workflow classes in their own space

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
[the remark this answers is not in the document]
djcooke: - If there is a Demultiplex_Stats file, can we skip the fastq_list file? Or read Demultiplex_Stats first and skip parts of fastq_list that are unnecessary?
- What is the difference between these and reason for preferring Demultiplex_Stats?
- Do they get written at different times? Do we always expect both?

It would be good to document the logic here

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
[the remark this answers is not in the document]
djcooke: The strings are hard-coded here, so I don't think it really makes a difference in that regard. It's better that Run Scanner output is consistent

avarsava:
> Having an option in the config to enable/disable it solves this

Using the 'novaseqx' or 'default' illumina processors is enabling or disabling per instrument. Enabling or disabling per run would be unsustainable, you'd have to edit the config for every run scanned.
As I mentioned, I am open to changing the novaseqx processor's name if we want to.

djcooke:
> Using the 'novaseqx' or 'default' illumina processors is enabling or disabling per instrument.

It would be simpler to just have a parameter within `DefaultIllumina` to enable or disable DRAGEN. You could avoid the deserialization issues (because there's no need for a separate DTO class either), the cloning (which has high potential to cause bugs/inconsistency if changes are made to IlluminaNotificationDto in the future), inconsistent `ObjectMapper`, etc. What advantage do you see in making this a completely separate processor?

> Enabling or disabling per run would be unsustainable

I agree, but it sounds like some runs will have analysis and some won't. It looks like you can tell from the samplesheet what workflows are expected, so this shouldn't be a problem.

avarsava:
> I agree, but it sounds like some runs will have analysis and some won't. It looks like you can tell from the samplesheet what workflows are expected, so this shouldn't be a problem.

This is already implemented. If the novaseqx/dragen/whatever-we-end-up-calling-it processor doesn't see any workflows in the sample sheet, then nothing gets added to expectedWorkflows and it immediately calls analysis done and doesn't do any waiting for analysis to appear.

avarsava:
Refactored to roll IlluminaDragenNotificationDto into IlluminaNotificationDto. Also added .ora support since there's two runs that were Unreadable without it.
It looks like BCLConvert prior to 4.1.7 does not generate a Manifest.tsv, so that's something we need to decide what to do about.
We also have one run that has DragenGermline and BCLConvert enabled in the samplesheet, but only DragenGermline output. The processor does the "right" thing and waits for BCLConvert output to appear, but it obviously never will. This is a waste of cycles and I've asked Illumina if this is the expected output when DragenGermline is enabled, but haven't heard back yet.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -0,0 +1,195 @@
...
+          try {
+            Analysis analysis = bclConvertAnalysis.get(filename);
+            for (AnalysisFile file : analysis.getFiles()) {
+              if (file.getPath().equals(filename)) {
+                file.setChecksum(manifestLine[1]);
+                break;
+              }
+            }
+          } catch (NullPointerException npe) {
+            log.info("Unable to map {} to an Analysis object", filename);
```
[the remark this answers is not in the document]
avarsava: The filename will contain the word 'Undetermined', but sure, i can amend it

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -0,0 +1,195 @@
...
+
+    // Check against samplesheet for okayness
+    isOk = true;
+    for (JsonNode item : samplesheet.getInfo().get("BCLConvert").get("Data")) {
+      Analysis analysisItem =
+          bclConvertAnalysis.get(
+              item.get("Sample_ID").textValue(),
+              item.get("Lane").textValue(),
+              item.get("Index").textValue(),
+              item.get("Index2").textValue());
```
callunity, line 158: I believe it's possible to do single-index sequencing. How would that affect this call?

callunity: Ooh, it's yet more complex: from [an Illumina page](https://knowledge.illumina.com/software/on-premises-software/software-on-premises-software-reference_material-list/000007337) talking about running DRAGEN with the first index read is a UMI and the second index read is an index:
```It is best practice to not leave an empty index column in the sample sheet (in this case an empty column for the first index). Some versions of BCL Convert will accept an empty column while others will error.```

Comment (avarsava) [reply scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java:158]: Are either of these things we will actually do though?

File: scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/Samplesheet.java [hunk 11]
```diff
@@ -0,0 +1,121 @@
...
+        }
+        switch (sectionName) {
+          case "[BCLConvert_Data]":
+            if (line[0].equals("Lane")) break; // Skip column label line
+            s++;
+            ObjectNode nested = mapper.createObjectNode();
+            nested.put("Lane", line[0]);
+            nested.put("Sample_ID", line[1]);
+            nested.put("Index", line[2]);
+            nested.put("Index2", line[3]); // Some have a line[4] like "Y27;I10;I10;Y27"
```
callunity, line 62: Why omit `line[4]`?

avarsava: Duplicated by BCLConvert.Settings.OverrideCycles

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
mlaszloffy: We're handling UMI's downstream of bclconvert, @lheisler I don't think any plans on changing that in the short/medium term? Anyway, it does look like depending on the run bases mask, index one may not be present. Will this set `Index` to null or fail?

Yes, we do single index sequencing, so index two (`Index2`) may be missing too. I'll send you an example run over slack.

lheisler: We will continue to manage UMIs downstream for now.  

Andre had built in some support with bcl2fastq to process UMIs upfront ie. generate fastq with UMIs extracted and recorded.  This was very specific to our UMI design, but we never made use of it.   I don't know what the current state of the Illumina tools and whether it can handle our UMI kit.  It would be worthwhile at some point to come investigate.  One early concern was releasing processed data for RUO projects, where they may have other/better approaches to using the UMIs.   As we've moved to move standard assays, i don't know that this is as much of a concern.   

Our current use of UMIs is that they are includesd in all WG library preps, which precedes capture for TarSeq.  (ie they are also in the TarSeq libraries).  For TarSeq they add value, and it is when these libraries are processed that they are extracted and used.  Data is small and this can be done quickly.   For WG libraries, they are just trimmed off, as they don't add as much value and take a long time to process.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/Analysis.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public void addFile(AnalysisFile file) {
+    files.add(file);
+  }
+
+  boolean isEmpty() {
+    return sample == null && lane == 0 && index == null && files.isEmpty();
+  }
+
+  ObjectNode toJson() {
```
djcooke, line 56: I'm guessing these are meant to be public?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/Analysis.java
```diff
@@ -0,0 +1,68 @@
...
+    ObjectNode analysisJson = DragenAnalysis.mapper.createObjectNode();
+    analysisJson.put("Sample", sample);
+    analysisJson.put("Lane", lane);
+    analysisJson.put("Index", index);
```
djcooke, line 60: - Doesn't match Run Scanner convention
- If there were proper models for everything, Jackson could handle all of this serialization automatically, significantly reducing the code size and improving consistency

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/Analysis.java
```diff
@@ -0,0 +1,68 @@
...
+    ArrayNode filesArray = DragenAnalysis.mapper.createArrayNode();
```
djcooke, line 61: You're depending on a weird static field of another class that hopefully was set..? None of these model objects should be holding onto an `ObjectMapper`. If they really need one, it should be passed into `toJson` or whatever method it's needed for, but they really shouldn't need it at all.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/Samplesheet.java
```diff
@@ -0,0 +1,121 @@
...
+          case "[BCLConvert_Settings]":
+            sampleSheetBCLConvertSettings.put(line[0], line[1]);
+            expectedWorkflows.put(DRAGENWorkflow.BCL_CONVERT, Boolean.FALSE);
+          case "[Cloud_Settings]": // Discard the cloud config
+          case "[Cloud_Data]":
+            break;
+          case "[Header]":
+          default:
+            if (line[0].startsWith("[")) continue; // Skip header lines we don't recognize
+            info.put(line[0], line[1]);
```
djcooke, line 74: This is just including any unrecognized line?

avarsava: Yes.

dto/src/test/java/ca/on/oicr/gsi/runscanner/rs/dto/test/AbstractNotificationDtoTest.java
```diff
@@ -33,10 +32,7 @@ public void testPartiallyPopulatedNotificationRoundTrip() throws Exception {
     notificationDto.setHealthType(HealthType.RUNNING);
 
     ObjectMapper mapper = new ObjectMapper();
-    mapper
-        .registerModule(new JavaTimeModule())
-        .setDateFormat(new ISO8601DateFormat())
-        .enable(SerializationFeature.INDENT_OUTPUT);
+    mapper.registerModule(new JavaTimeModule()).enable(SerializationFeature.INDENT_OUTPUT);
```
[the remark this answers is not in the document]
avarsava: The ISO8601DateFormat class was marked as deprecated and also doesn't match the code it's testing since those two lines were the only usages of ISO8601DateFormat in the codebase.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/AnalysisFile.java
```diff
@@ -0,0 +1,85 @@
...
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+// Represents one file output by DRAGEN
+public class AnalysisFile {
+  private Path path;
+  private String checksum;
```
avarsava, line 12: rename to indicate crc32

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/samplesheet/Samplesheet.java
```diff
@@ -0,0 +1,146 @@
...
+          continue;
+        }
+        switch (sectionName) {
+          case "BCLConvert_Data":
+            if (line[0].equals("Lane")) break; // Skip column label line
+            bclConvertSection = (SamplesheetBCLConvertSection) getByName("BCLConvert");
+            if (bclConvertSection == null) {
+              bclConvertSection = new SamplesheetBCLConvertSection();
+            }
+            bclConvertSection.addDatum(line[0], line[1], line[2], line[3]);
```
avarsava, line 80: handle changes in order

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/samplesheet/Samplesheet.java
```diff
@@ -0,0 +1,146 @@
...
+            if (bclConvertSection == null) {
+              bclConvertSection = new SamplesheetBCLConvertSection();
+            }
+            bclConvertSection.addSetting(line[0], line[1]);
+            addToSamplesheet(bclConvertSection);
+            expectedWorkflows.put(DRAGENWorkflow.BCL_CONVERT, Boolean.FALSE);
+          case "Cloud_Settings": // Discard the cloud config
+          case "Cloud_Data":
+            break;
+          default:
```
avarsava, line 95: discard unsupported sections, also the header and reads

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/IlluminaNotificationDto.java
```diff
@@ -22,6 +33,18 @@ public class IlluminaNotificationDto extends NotificationDto {
   private String workflowType;
   private IndexSequencing indexSequencing;
 
+  // @JsonDeserialize(using = JsonNullDeserializer.class)
+  private Analysis analysis;
```
avarsava, line 37: handle >1 Analysis (ie reattempts)

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/samplesheet/Samplesheet.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Samplesheet {
```
avarsava, line 19: split into struct and processor

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/IlluminaNotificationDto.java
```diff
@@ -22,6 +25,17 @@ public class IlluminaNotificationDto extends NotificationDto {
   private String workflowType;
   private IndexSequencing indexSequencing;
 
+  public Map<Integer, Analysis> analysis = new HashMap<>();
```
djcooke, line 28: I'd expect this to serialize something like this:

```
"analysis": {
  1: { ... (Analysis object) },
  2: { ... (Analysis object) }
}
```

That's not valid JSON (integer keys), so I'm guessing that's the serialization issue, though I'm surprised it doesn't throw an exception.

A possible good solution is to store the attempt number as a field in `Analysis` and make `IlluminaNotificationDto.analysis` a `List<Analysis>` or similar `Set`/array. This would also give a purpose to the `Analysis` interface, which right now is basically just `Object`. Could also make `Analysis` an abstract class instead of an interface.

A worse solution would be to just change the `Map` key to `String` and store the number as a string.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/BCLConvert.java
```diff
@@ -0,0 +1,197 @@
...
+  public boolean isOk() {
+    return isOk;
+  }
+
+  public DragenWorkflowAnalysis getResult() throws IOException {
+    if (result == null) process();
+    return result;
+  }
+
+  private void process() throws IOException {
```
djcooke, line 38: Dtos module should not contain any processing

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenAnalysis.java
```diff
@@ -0,0 +1,28 @@
+package ca.on.oicr.gsi.runscanner.dto.dragen;
+
+import ca.on.oicr.gsi.runscanner.dto.Analysis;
+import ca.on.oicr.gsi.runscanner.dto.dragen.samplesheet.Samplesheet;
+import java.util.HashMap;
+import java.util.Map;
+
+// Represents an entire attempt at a DRAGEN Pipeline
+public class DragenAnalysis implements Analysis {
+  Samplesheet samplesheet;
```
djcooke, line 10: The analysis attempt has a samplesheet and BCLConvert has a different samplesheet?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenAnalysis.java
```diff
@@ -0,0 +1,28 @@
...
+  Map<String, DragenWorkflowAnalysis> analyses = new HashMap<>();
```
djcooke, line 11: This is another field that could be defined in the Analysis interface. I imagine all workflow analysis should have some common attributes - start and completion time? There could be a `WorkflowAnalysis` abstract class that defines these, and `Analysis` could take a type parameter to specify the implementation. e.g.

```
public abstract class Analysis<T extends WorkflowAnalysis>

public class DragenAnalysis extends Analysis<DragenWorkflowAnalysis>
```

I think the workflow name also belongs in the WorkflowAnalysis class. Standardizing as many fields as possible will make for a cleaner API that's easier to understand and work with.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenAnalysisUnit.java
```diff
@@ -0,0 +1,53 @@
...
+    this.lane = i;
+  }
+
+  public void setIndex(String str) {
+    this.index = str;
+  }
+
+  public void setIndex(String index1, String index2) {
+    this.setIndex(new StringBuilder(index1).append("-").append(index2).toString());
+  }
```
djcooke, line 44: Why combine the indices into one string?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenWorkflowAnalysis.java
```diff
@@ -0,0 +1,76 @@
...
+
+  public DragenAnalysisUnit get(Path filePath) {
+    List<DragenAnalysisUnit> list =
+        analyses
+            .stream()
+            .filter(a -> a.getFiles().stream().anyMatch(f -> f.getPath().equals(filePath)))
+            .toList();
+    if (list.size() > 1) {
+      throw new IllegalStateException(
+          "Can't have more than one Analysis with same file " + filePath);
```
djcooke, line 41: This is allowing bad data to be added and then validating it later upon retrieval? Or not validating it at all if get isn't called after adding?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenWorkflowAnalysis.java
```diff
@@ -0,0 +1,76 @@
...
+    return get(sample, Integer.parseInt(lane), index);
+  }
+
+  public DragenAnalysisUnit get(String sample, int lane, String index) {
+    return analyses
+        .stream()
+        .filter(a -> a.getSample().equals(sample) && a.getLane() == lane)
+        .findFirst()
+        .orElse(new DragenAnalysisUnit());
+  }
```
djcooke, line 31: This is a "get" method, but it also creates something new, empty, and invalid if the thing requested doesn't already exist. Very odd for a DTO and not something an API consumer would expect. If this logic is only here for processing purposes, it should be in the processor.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/samplesheet/Samplesheet.java
```diff
@@ -0,0 +1,42 @@
+package ca.on.oicr.gsi.runscanner.dto.dragen.samplesheet;
+
+import java.time.Instant;
+import java.util.LinkedList;
+import java.util.List;
+
+public class Samplesheet {
+  private List<SamplesheetSection> info;
+  private Instant mtime;
```
djcooke, line 9: Please use similar naming for all these time fields.

- `AnalysisFile` has `created` and `modified`
- `DragenWorkflowAnalysis` has `startTime` and `completionTime`
- Here is `mtime`

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/IlluminaNotificationDto.java
avarsava: Given that I am already getting the number as a String and using Integer.valueOf() to parse it at NovaseqXProcessor:50, I will just stop doing that and use Strings

avarsava: Unfortunately making this change did not fix the serialization problem

dto/pom.xml
```diff
@@ -34,6 +32,14 @@
...
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>ca.on.oicr.gsi</groupId>
+      <artifactId>server-utils</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
```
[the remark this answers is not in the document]
avarsava: Pair comes from server-utils. logging can go

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/IlluminaNotificationDto.java
avarsava: also, even with an integer key, that part always serialized successfully, it's the value that doesn't:
image

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/BCLConvert.java
avarsava: Moved BCLConvert back to the processor package and made it return a dto instead

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenAnalysis.java
avarsava: Done

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenWorkflowAnalysis.java
avarsava: Done.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenWorkflowAnalysis.java
avarsava: done

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenAnalysisUnit.java
avarsava: done

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenWorkflowAnalysis.java
```diff
@@ -0,0 +1,65 @@
...
+
+  public DragenAnalysisUnit get(String sample, String lane, String index) {
+    return get(sample, Integer.parseInt(lane), index);
+  }
+
+  public DragenAnalysisUnit get(String sample, int lane, String index) {
+    return analyses
+        .stream()
+        .filter(a -> a.getSample().equals(sample) && a.getLane() == lane)
```
callunity, line 30: Why no filter on index?

avarsava: Because I forgot! fixed

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/samplesheet/Samplesheet.java
avarsava: Made these a bit more consistent.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenWorkflowAnalysis.java
```diff
@@ -0,0 +1,66 @@
...
+      return null;
+    }
+    return list.get(0);
+  }
+
+  public void put(DragenAnalysisUnit newDragenAnalysisUnit) {
+    for (AnalysisFile af : newDragenAnalysisUnit.getFiles()) {
+      get(af.getPath()); // throw out the result, we just want the file path validation
+    }
+    DragenAnalysisUnit oldDragenAnalysisUnit =
```
callunity, line 56: What do you think about `existingDragenAnalysisUnit` and `updatedDragenAnalysisUnit` or `currentDragenAnalysisUnit`, for clarity?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -0,0 +1,190 @@
...
+        for (String[] manifestLine : manifestLines) {
+          // 0 = path, 1 = crc32 checksum
+          Path filename = new File(rootDir, manifestLine[0]).toPath();
+
+          // When this is null, it's often for an Undetermined read. We do not care.
+          try {
+            DragenAnalysisUnit dragenAnalysisUnit = bclConvertAnalysis.get(filename);
+            for (AnalysisFile file : dragenAnalysisUnit.getFiles()) {
+              if (file.getPath().equals(filename)) {
+                file.setCrc32Checksum(manifestLine[1]);
```
avarsava, line 95: do we need put() them back

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -0,0 +1,190 @@
...
+
+        for (String[] demuxLine : demultiplexLines) {
+          if (demuxLine[0].startsWith("Lane")) continue; // skip the column labels
+          // 0 = Lane, 1 = SampleId, 2 = Index, 3= # Reads, 4 = # Perfect Index Reads,
+          // 5 = # One Mismatch Index Reads, 6 = # Two Mismatch Index Reads, 7 = % Reads,
+          // 8 = % Perfect Index Reads, 9 = % One Mismatch Index Reads,
+          // 10 = % Two Mismatch Index Reads
+          DragenAnalysisUnit dragenAnalysisUnit =
+              bclConvertAnalysis.get(demuxLine[1], demuxLine[0], demuxLine[2]);
+          if (dragenAnalysisUnit == null) dragenAnalysisUnit = new DragenAnalysisUnit();
```
avarsava, line 134: should probably be an err

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -0,0 +1,190 @@
...
+        break;
+      } else {
+        for (AnalysisFile itemFile : dragenAnalysisUnitItem.getFiles()) {
+          if (itemFile == null
+              || itemFile.getPath() == null
+              || itemFile.getPath().toString().isBlank()
+              || itemFile.getInfo() == null
+              || itemFile.getCrc32Checksum() == null
+              || itemFile.getCrc32Checksum().isBlank()) {
+            isOk = false;
```
avarsava, line 165: also don't add the unit if validation fails

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/Analysis.java
```diff
@@ -0,0 +1,16 @@
+package ca.on.oicr.gsi.runscanner.dto;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public abstract class Analysis<T extends WorkflowAnalysis> {
```
avarsava, line 6: @JsonTypeInfo

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/Analysis.java
```diff
@@ -0,0 +1,16 @@
...
+  Map<String, T> analyses = new HashMap<>();
```
avarsava, line 7: make property of workflowanalysis

dto/pom.xml
```diff
@@ -34,6 +32,10 @@
       <artifactId>hamcrest-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>ca.on.oicr.gsi</groupId>
+      <artifactId>server-utils</artifactId>
```
avarsava, line 37: just copy pair in

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
```diff
@@ -0,0 +1,197 @@
...
+            return dto;
+          }
+
+          dragenAnalysis = new DragenAnalysis(samplesheet);
+
+          if (isWorkflowExpected(DragenWorkflow.BCL_CONVERT)) {
+            BCLConvert bclConvert = new BCLConvert();
+            dragenAnalysis.put("BCLConvert", bclConvert.process(samplesheet, analysisAttempt));
+
+            // TODO: Should we not put() anything if not OK?
```
avarsava, line 68: A: no

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenWorkflowAnalysis.java
avarsava: Actually, in the case of a truly 'new' DragenAnalysisUnit (there is no old unit to remove), then it is not 'updated', but 'current' and 'existing' could seem like they mean the same thing (the existing unit is the current one until it is updated). I do think 'old' and 'new' are clear

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/NovaseqXProcessor.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.TimeZone;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NovaseqXProcessor extends DefaultIllumina {
+  private Map<DragenWorkflow, Boolean> expectedWorkflows = new HashMap<>();
```
avarsava, line 26: should be related to each attempt

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenWorkflowAnalysis.java
callunity: Ok, thanks for clarifying

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/Analysis.java
avarsava: @djcooke I did this, and it writes out the property name i gave it, but only that, any ideas?
image

avarsava: Update as of [f72bb63](https://github.com/miso-lims/runscanner/pull/165/commits/f72bb632cb7e5e976c786fd14792f1c43218b28a) there's also the attempt number
image

callunity: The SampleSheetSection and its implementation might also need Jackson annotations?

callunity: Jackson might also need explicit getters - `Analysis.analyses` and `DragenAnalysis.samplesheet` don't yet have getters

djcooke: Writing unit tests would be the easiest way to troubleshoot and much faster than scanning a run directory to test every change. Start with the smallest unit and work your way out until a full DragenAnalysis is serializing and deserializing fine. Cardea has similar tests: https://github.com/oicr-gsi/cardea/blob/main/cardea-data/src/test/java/ca/on/oicr/gsi/cardea/JacksonTest.java

avarsava: Thanks Heather, this at least got the samplesheet writing out. 
So if everything needs a getter for serialization, I imagine it also needs a setter for deserialization? How do you keep someone from modifying something they shouldn't after the fact then? For example, the samplesheet shouldn't be able to be overwritten once it's been parsed in.

callunity: It looks like Jackson uses reflection to access private and protected properties. I've seen (but not verified) one article which claims Jackson will use reflection to populate a field if it has a getter. Writing the unit tests like Dillan suggested above will be a quick way to explore this!

djcooke: Immutability would prevent modifying the object, but Run Scanner's DTOs are not immutable

avarsava: as of https://github.com/miso-lims/runscanner/pull/165/commits/cb14227432607ce152046722836f1ee1ba9c2463: Getting this error on the DragenWorkflowAnalysis serialization test:
`com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `ca.on.oicr.gsi.runscanner.dto.dragen.DragenWorkflowAnalysis` (although at least one Creator exists): cannot deserialize from Object value (no delegate- or property-based Creator)`

Not sure why, because WorkflowAnalysis has the correct JsonTypeInfo annotation?

djcooke: It might actually be fine if you rename the parameter to `workflowName` so that it matches the field name it's supposed to be used for

avarsava: Unfortunately, renaming it didn't help

avarsava: Does there need to be? The annotation alone makes it serialize with both fields set correctly:

`{"suite":"DRAGEN","completionTime":31556889864403199.999999999,"startTime":-31557014167219200.000000000,"workflowName":"TEST","analyses":[{"sample":"TEST_SAMPLE","lane":1,"index":"AAAAAA-TTTTTT","files":[{"path":"file:///path/to/file","crc32Checksum":"abcdefg","size":1000,"createdTime":0.0,"modifiedTime":0.0,"info":{}}]}],"suite":"DRAGEN"}`

I would think the annotation would tell jackson to look at 'suite' at the correct level and construct the class indicated in the @Type annotation?

callunity: Why is the constructor parameter called `name` when the WorkflowAnalysis field it maps to is `workflowName`?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/WorkflowAnalysis.java
```diff
@@ -0,0 +1,50 @@
...
+  public Instant getStartTime() {
+    return startTime;
+  }
+
+  public void setStartTime(Instant i) {
+    this.startTime = i;
+  }
+
+  public String getSuite() {
+    return "DRAGEN";
```
callunity, line 48: Is this intended to be constant for `WorkflowAnalysis`?

avarsava: No, leftover from trying to get it to deserialize

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/Analysis.java
avarsava: Finished all the unit tests. All passing. Still doesn't actually work when running runscanner

dto/src/test/java/ca/on/oicr/gsi/runscanner/rs/dto/test/dragen/JacksonTest.java
```diff
@@ -0,0 +1,189 @@
...
+import ca.on.oicr.gsi.runscanner.dto.dragen.samplesheet.Samplesheet;
+import ca.on.oicr.gsi.runscanner.dto.dragen.samplesheet.SamplesheetBCLConvertSection;
+import ca.on.oicr.gsi.runscanner.dto.dragen.samplesheet.SamplesheetBCLConvertSection.SamplesheetBCLConvertDataEntry;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import java.nio.file.Path;
+import java.time.Instant;
+import org.junit.jupiter.api.Test;
+
+// TODO this isn't running on build
```
avarsava, line 18: I thought, 'maybe an older version of surefire doesn't look in subdirectories'. So I tried upgrading surefire to 3.5.2. That version _only_ runs JacksonTest and skips the NotificationDto tests one level up. ???

avarsava:
Will roll back HEAD^ and rebase on master once #173 is merged.
I don't understand why the illumina processor test is failing, works on my machine

avarsava:
> Will roll back HEAD^ and rebase on master once #173 is merged. I don't understand why the illumina processor test is failing, works on my machine

The problem is that the path for the AnalysisFiles in the test end up with a path of `path=runscanner/runscanner/scanner/target/test-classes/illumina/20240531_LH00130_0061_A22CC55LT4/Analysis/2/Data/BCLConvert/fastq/STSMOH_01548_01_LB01-01_S1_L001_R1_001.fastq.gz` (two /runscanner/s, which is not how they are in the reference)

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
```diff
@@ -443,11 +457,12 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
           Path baseCallDirectory =
               Paths.get(dto.getSequencerFolderPath(), "Data", "Intensities", "BaseCalls");
           // Check that each lane directory is complete
+          IlluminaNotificationDto finalDto = dto;
           boolean dataCopied =
               IntStream.rangeClosed(1, dto.getLaneCount()) //
                   .mapToObj(lane -> String.format("L%03d", lane)) //
                   .map(baseCallDirectory::resolve) //
-                  .allMatch(laneDir -> isLaneComplete(laneDir, dto));
+                  .allMatch(laneDir -> isLaneComplete(laneDir, finalDto));
```
djcooke, line 465: ?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/AnalysisFile.java
```diff
@@ -0,0 +1,101 @@
...
+// Represents one file output by DRAGEN
+public class AnalysisFile {
+  private Path path;
+  private String crc32Checksum;
+  private long size;
+  private Instant createdTime;
+  private Instant modifiedTime;
+
+  // What will be in this map will be unique to the particular DRAGEN Workflow
+  private Map<String, Object> info = new HashMap<>();
```
djcooke, line 18: Jackson won't be able to deserialize `Object` and there should be strictly defined fields anyway. Any workflow that needs its own fields should have a subclass of AnalysisFile. Looks like BCLConvert needs read count and read number which are currently being added to this as "read_count" and "read_number" in `ProcessDragen`, which is also inconsistent formatting with the rest of Run Scanner's output

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenPipelineRun.java
```diff
@@ -0,0 +1,38 @@
+package ca.on.oicr.gsi.runscanner.dto.dragen;
+
+import ca.on.oicr.gsi.runscanner.dto.PipelineRun;
+import ca.on.oicr.gsi.runscanner.dto.dragen.samplesheet.Samplesheet;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Objects;
+
+// Represents an entire attempt at a DRAGEN Pipeline
+public class DragenPipelineRun extends PipelineRun<DragenWorkflowRun> {
+  Samplesheet samplesheet;
```
djcooke, line 10: I understand the information stored in this `Samplesheet` object is needed during processing to check whether the workflows are complete, but should it be included in the output? It's mostly duplicating the information in `DragenAnalysisUnit`

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/DragenAnalysisUnit.java
```diff
@@ -0,0 +1,88 @@
...
+    this.index = str;
+  }
+
+  public void setIndex(String index1, String index2) {
+    if (index2 == null) {
+      this.setIndex(index1);
+    } else {
+      this.setIndex(new StringBuilder(index1).append("-").append(index2).toString());
+    }
+  }
```
djcooke, line 50: Output index 1 and 2 separately. If something downstream wants to combine them, it can easily do that. No reason to put the worse format in Run Scanner

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/dragen/samplesheet/SamplesheetBCLConvertSection.java
```diff
@@ -0,0 +1,129 @@
...
+    private final String lane;
+    private final String sampleId;
+    private final String index;
+    private final String index2;
+
+    public SamplesheetBCLConvertDataEntry(
+        @JsonProperty("lane") String lane,
+        @JsonProperty("sampleId") String sampleId,
+        @JsonProperty("index") String index,
+        @JsonProperty("index2") String index2) {
```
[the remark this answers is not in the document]
avarsava: Without these annotations, the JacksonTests fail with "com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `ca.on.oicr.gsi.runscanner.dto.dragen.samplesheet.SamplesheetBCLConvertSection$SamplesheetBCLConvertDataEntry` (no Creators, like default constructor, exist): cannot deserialize from Object value (no delegate- or property-based Creator)". With the annotations, the tests pass.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java
avarsava: Without this, build fails with
"[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project scanner: Compilation failure
[ERROR] /home/avarsava/Workspace/runscanner/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultIllumina.java:[464,64] local variables referenced from a lambda expression must be final or effectively final"

avarsava:
> The difference is the rest of the project uses junit 4 and you added and used junit 5 (junit-jupiter) only here

I don't know why github won't let me reply to this, but thank you, I wouldn't have caught that myself

scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/IlluminaProcessorTest.java
```diff
@@ -23,8 +23,8 @@ protected NotificationDto process(File directory) throws IOException {
 
   @Override
   public void testGoldens() throws IOException {
-    if (!System.getProperty("skipIllumina", "true").equals("true")) {
-      checkDirectory("/illumina");
-    }
+    // if (!System.getProperty("skipIllumina", "true").equals("true")) {
```
callunity, line 26: Looks like this might have stayed commented out by accident?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -0,0 +1,266 @@
...
+    // Check against samplesheet for okayness
+    boolean isOk = true;
+    for (SamplesheetBCLConvertDataEntry item :
+        ((SamplesheetBCLConvertSection) samplesheet.getByName("BCLConvert")).getData()) {
+      DragenAnalysisUnit dragenAnalysisUnitItem =
+          bclConvertWorkflowRun.get(
+              item.getSampleId(),
+              item.getLane(),
+              item.getIndex(),
+              reverseComplement(item.getIndex2()));
```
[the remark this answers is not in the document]
avarsava: RunInfo.xml is about the sequencing, not the analysis. The settings for the two can be distinct. 
We discussed this back in December, did I misunderstand?
https://oicr.slack.com/archives/GJNEQ8LES/p1733516523088469

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/ProcessDragen.java
```diff
@@ -0,0 +1,103 @@
...
+          dragenPipelineRun = new DragenPipelineRun(attemptNum);
+
+          // Exit early if we didn't find any workflows we support
+          if (samplesheet.noneExpected()) {
+            dragenPipelineRun.setPipelineStatus(PipelineStatus.UNSUPPORTED);
+            dto.addPipelineRun(dragenPipelineRun);
+            continue;
+          }
+
+          // Exit early if Samplesheet finds a SoftwareVersion < 4.1.7, our minimum supported
```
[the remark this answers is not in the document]
avarsava: I will write documentation for this feature once the code is approved, it's just in a lot of flux even now

scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/IlluminaProcessorTest.java
avarsava: Couldn't get this setting to work in intellij, will uncomment these lines once the code is approved.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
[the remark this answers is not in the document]
callunity: That link says that the NextSeq 2000 and the NovaSeq X/X Plus follow the same pattern of "enter data in forward orientation and BCLConvert will reverse complement it automatically", but that the NovaSeq 6000 and NextSeq 550 (which can run BCLConvert) need it to not be reverse complemented here as these instruments are supposed to have indices entered as the reverse complement in the samplesheet. 

Given Illumina's inconsistency on this count I think it would be very reasonable to decide we're only supporting DRAGEN parsing for the NextSeq 1000/2000 and NovaSeq X/X Plus at this time rather than try to cover use cases for instruments we are not running, but that should be commented in the code here and added to the release notes.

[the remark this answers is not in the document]
avarsava: As mentioned, I will write documentation for this feature once the code is approved. I have scoped this development strictly to the X+ and will mention that in the documentation.

callunity (approved):

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -0,0 +1,266 @@
...
+      }
+    }
+    if (isOk) bclConvertWorkflowRun.complete();
+    return bclConvertWorkflowRun;
+  }
+
+  private static AnalysisFile fastqFromFilename(File rootDir, String fileName, int readNumber)
+      throws IOException {
+    Path fullPath = Paths.get(rootDir.getPath(), "/Data/BCLConvert/fastq/", fileName);
+    if (!Files.exists(fullPath)) {
```
djcooke, line 200: An empty string (and probably null) for `fileName` would result in checking that the `fastq` directory exists. Should probably check for null and empty string at the start and return null

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -0,0 +1,266 @@
...
+            file2 = fastqFromFilename(rootDir, fastq[5], 2);
+
+        if (file1 != null && file1.getModifiedTime().compareTo(max_date) > 0)
+          max_date = file1.getModifiedTime();
+
+        if (file2 != null && file2.getModifiedTime().compareTo(max_date) > 0)
+          max_date = file2.getModifiedTime();
+
+        dragenAnalysisUnit.addFile(file1);
+        dragenAnalysisUnit.addFile(file2);
```
djcooke, line 71: Don't add if null

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -0,0 +1,266 @@
...
+            DragenAnalysisUnit dragenAnalysisUnit = bclConvertWorkflowRun.get(filename);
+            for (AnalysisFile file : dragenAnalysisUnit.getFiles()) {
+              if (file.getPath().equals(filename)) {
+                file.setCrc32Checksum(manifestLine[1]);
+                break;
+              }
+            }
+            bclConvertWorkflowRun.put(dragenAnalysisUnit);
+          } catch (NullPointerException npe) {
+            log.info("Unable to map {} to an Analysis object", filename);
```
djcooke, line 102: Check for null instead of handling NPE

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/samplesheet/Samplesheet.java
```diff
@@ -0,0 +1,241 @@
...
+              .filter(line -> !(line.length == 0 || line.length == 1 && line[0].isEmpty()))
+              .toList();
+
+      SamplesheetBCLConvertSection bclConvertSection = null;
+      SamplesheetReadsSection readsSection = null;
+      Matcher headerMatcher;
+      String sectionName = "";
+      boolean bclDataFirstLine = true;
+      Map<String, Integer> lineIndices = new HashMap<>();
+      for (String[] line : lines) {
```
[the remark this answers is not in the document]
avarsava: There's already an attempt to filter out blank lines in the stream that builds the lines array, I can expand it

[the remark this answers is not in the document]
djcooke: Ah, I somehow missed that. Looks good

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/ProcessDragen.java
```diff
@@ -35,14 +34,21 @@ public static IlluminaNotificationDto analyse(
       for (File analysisAttempt : Objects.requireNonNull(analysisDir.listFiles())) {
         if (analysisAttempt.isDirectory() && analysisAttempt.getName().matches(NUMERAL)) {
           int attemptNum = Integer.parseInt(analysisAttempt.getName());
-          Samplesheet samplesheet = new Samplesheet(analysisAttempt);
+          dragenPipelineRun = new DragenPipelineRun(attemptNum);
+          Samplesheet samplesheet;
+          try {
+            samplesheet = new Samplesheet(analysisAttempt);
+          } catch (IllegalStateException ise) {
```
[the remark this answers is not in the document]
avarsava: The constructor throws this exception at expected failure points, and those are handled here, to set the SCAN_ERROR status we agreed to. I'm not sure I understand what alternative you're suggesting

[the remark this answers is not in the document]
djcooke: IllegalStateException is a runtime (unchecked) exception, which means you need to know to expect it and catch it if you want to handle it. This is because they're not really meant to be caught/handled - they are meant to represent programming errors or unrecoverable conditions.

I'm suggesting that the constructor should be throwing some type of checked exception (e.g. IOException) because these are conditions that you can handle (as you are) and should be forced to handle, so someone can't decide to remove this try/catch in the future without realizing.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/BCLConvert.java
```diff
@@ -23,12 +26,36 @@ public class BCLConvert {
...
+        Instant instantDateTime =
+            LocalDateTime.parse(event[0], dateTimeFormatter).atOffset(ZoneOffset.UTC).toInstant();
+        if (start) {
+          bclConvertWorkflowRun.setStartTime(instantDateTime);
+          start = false;
+        }
+        bclConvertWorkflowRun.setCompletionTime(instantDateTime);
+      }
+    } else {
+      log.info("No b2c_dragen_events.csv for {}", rootDir);
```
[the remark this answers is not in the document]
avarsava: lack of start/completion time will trip here https://github.com/miso-lims/runscanner/pull/165/files#diff-7c019d45586155cfe9a93816df3e576db6d958f02d4d68c05600d907da8acd9bR193-R196

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/dragen/ProcessDragen.java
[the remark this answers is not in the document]
avarsava: Sorry, I still don't understand. Both exceptions are in the constructors signature: https://github.com/miso-lims/runscanner/pull/165/files#diff-ebd5366087e5ffb50492bf4b0624828efb158ab4d86e4f92a8289a068eedb47aR47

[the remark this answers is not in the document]
djcooke: RuntimeExceptions may show a warning in your IDE, but checked exceptions cause a compile error if not handled. Adding it to the method signature does not make it a checked exception. IOException may not make sense for this case, but that was just an example of a checked exception. There are others, and you can also create your own.

https://docs.oracle.com/javase/tutorial/essential/exceptions/runtime.html

djcooke (approved):

avarsava:
Time to write docs

avarsava:
@mlaszloffy @rnaioicr only requesting your review for the documentation commit: https://github.com/miso-lims/runscanner/pull/165/commits/eee6d9189ad9626430a5baca91517471554e2e47

mlaszloffy (approved):

docs/appendices.md
```diff
@@ -302,3 +318,37 @@ unsupported.
...
+Run Scanner parses a number of files out of the run directory in order to serve analysis data.
+For compatibility between versions, sometimes a number of file locations are checked for each file.
+Newer versions of an analysis suite may change the location of a file. If you detect that this has
+occured, please
+[open a GitHub Issue](https://github.com/miso-lims/runscanner/issues/new) detailing the situation
+and including the file location.
+
+### Illumina DRAGEN
+#### BCLConvert
+Let `attempt` be `sequencer location/run directory/Analysis/attempt number/`
```
djcooke, line 340: Blank lines after each heading

docs/troubleshooting.md
```diff
@@ -1,3 +1,4 @@
+## Sequencing Output
```
djcooke, line 1: Blank lines after all headings in this file

docs/troubleshooting.md
```diff
@@ -100,3 +101,44 @@ If runs from an Illumina sequencer are universally failing, ensure that
...
+unsupported by Run Scanner, or the DRAGEN version indicated in the samplesheet is lower than
+Run Scanner's minimum supported version (4.1.7), Run Scanner sets `pipelineStatus` to `UNSUPPORTED`.
+Please see [Illumina DRAGEN](./analysis.md) for the current list of DRAGEN workflows supported
+by Run Scanner.
+
+#### Pipeline run has a `pipelineStatus` of `SCAN_ERROR`
+A supported workflow has the correct files, however, there is some malformation in the file content
+such that Run Scanner is unable to parse the file. This is an error which requires developer 
+attention. 
+There should be an error message in your tomcat logs that looks like "SCAN_ERROR in [filename] for
```
djcooke, line 135: Need a blank line if you want this to be a new paragraph

docs/troubleshooting.md
```diff
@@ -100,3 +101,44 @@ If runs from an Illumina sequencer are universally failing, ensure that
...
+[directory]". If you are able, please
+[open a GitHub Issue](https://github.com/miso-lims/runscanner/issues/new) detailing the situation
+and including the file contents.
+
+#### Workflow run has a status of `FAILED`
+Run Scanner will set a workflow run to `FAILED` if Run Scanner is unable to locate all of the 
+information it requires to serve analysis data, but DRAGEN has indicated via touchfile that
```
djcooke, line 142: This matches the description of `PENDING` status from the "About Workflow Run Status," where `FAILED` is described as "Run Scanner has evidence of a failure of the workflow run."

docs/analysis.md
```diff
@@ -0,0 +1,95 @@
...
+See below for platform-specific information. If a runs `analysisExpected` field is `false` and you 
+are expecting analysis information for that run, please see 
+[Troubleshooting Analysis Output](./troubleshooting.md).
+
+If disabled, the `analysisExpected` field of the /run endpoint will always be set to `false`,
+and the `pipelineRuns` field of the /run endpoint will always be an empty array.
+
+## About Pipeline Status
+
+The `pipelineRuns` field of the /run endpoint is an array of data structures specific to the
```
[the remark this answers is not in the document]
avarsava: Where would a user retrieve this data if not the run endpoint?

docs/troubleshooting.md
avarsava: PENDING is "Run Scanner has not found all of the information it needs to serve analysis data yet", FAILED is "Run Scanner is unable to locate all of the information it requires to serve analysis data, **but DRAGEN has indicated via touchfile that it is done writing out analysis data**". The implication is that the missing data can never appear. How would you recommend I clarify this?

docs/analysis.md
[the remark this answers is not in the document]
djcooke: There are 3 REST API endpoints that serve full run data for one or more runs. See the API docs. "/run" isn't really even an endpoint, though "/run/{name}" is.

docs/troubleshooting.md
djcooke: `FAILED` means Run Scanner thinks the workflow is failed, which seems obvious without explanation, but becomes unclear when reading this.

One reason it might be marked `FAILED` when that is incorrect is that required files or permissions are missing. It may be the only reason for now, but that's likely to change, and there's a lot of documentation here that you're assuming will be accurately maintained with every future change.

docs/troubleshooting.md
avarsava: Why? Those headers are old and seem to be rendering properly: https://miso-lims.readthedocs.io/projects/runscanner/en/latest/troubleshooting/

rnaioicr (approved):

PR #171, opened 2024-12-17
[GLT-4278] Grab Revio run information

Christopher-Ho:
JIRA Ticket: https://jira.oicr.on.ca/browse/GLT-4278

Add initial support for PacBio Revio by reading in a Revio run directory and pulling information about the run.

### DIFF
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
@@ -20,7 +20,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
-import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Scanner;
 import java.util.TimeZone;
@@ -334,8 +333,7 @@ public class RevioPacBioProcessor extends RunProcessor {
    * Filter runDirectory for Transfer_Test files and grab earliest creation time
    *
    * @param runDirectory which we are currently processing
-   * @return earliest file creation time
-   * @throws NoSuchElementException if the .Transfer_Test files are missing
+   * @return earliest file creation time or null if no Transfer_Test files are found
    */
   private Instant startTimeFromTransferTest(File runDirectory) throws IOException {
     List<File> TransferTestFiles =
@@ -360,8 +358,7 @@ public class RevioPacBioProcessor extends RunProcessor {
    * Filter runDirectory for .Transferdone file and grab latest creation time
    *
    * @param runDirectory which we are currently processing
-   * @return latest file creation time
-   * @throws NoSuchElementException if the .Transferdone files are missing
+   * @return latest file creation time or null if no Transferdone files are found
    */
   private Instant completionTimeFromTransferDone(File runDirectory) throws IOException {
     List<File> TransferDoneFiles =


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,335 @@
...
+
+    // Get the number of SMRT Cells in the run directory
+    dto.setLaneCount((int) getCellDirectory(runDirectory).count());
+
+    // Presence of Transfer_Test_*.txt indicates the run has started, get the creation time
+    getMetadataDirectory(runDirectory)
+        .flatMap(
+            metadataDirectory ->
+                Arrays.stream(
+                    metadataDirectory.listFiles(file -> TRANSFER_TEST.test(file.getName()))))
```
callunity, line 195: I think I'm missing something here, but are you trying to get a list of files and then filter them based on whether they match the TRANSFER_TEST pattern? And if so, would you consider doing it as a separate `flatMap(get list of files).filter(TRANSFER_TEST.test(...))` for clarity?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,335 @@
...
+                throw new RuntimeException(e);
+              }
+            });
+
+    // We are good to process this run
+    getMetadataDirectory(runDirectory)
+        .flatMap(
+            metadataDirectory ->
+                Arrays.stream(
+                    metadataDirectory.listFiles(file -> file.getName().endsWith(".metadata.xml"))))
```
callunity, line 216: Could use a `filter` here as well

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,335 @@
...
+    // We create one DTO for a run, but there are going to be many wells with
+    // independent and
+    // duplicate metadata that will simply
+    // overwrite in the shared DTO. If the data differs, the last well wins.
+    PacBioNotificationDto dto = new PacBioNotificationDto();
+    dto.setPairedEndRun(false);
+    dto.setSequencerFolderPath(runDirectory.getAbsolutePath());
+
+    // Get the number of SMRT Cells in the run directory
+    dto.setLaneCount((int) getCellDirectory(runDirectory).count());
```
callunity, line 188: getCellDirectory(runDirectory) is only being used in this file for the value of `.count()`. What do you think of saving the count to a variable up here and using the variable in lieu of the `getCellDirectory(runDirectory).count()` calls throughout?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,335 @@
...
+          .flatMap(
+              metadataDirectory ->
+                  Arrays.stream(
+                      metadataDirectory.listFiles(file -> TRANSFER_DONE.test(file.getName()))))
+          .forEach(
+              s -> {
+                try {
+                  FileTime fileTime =
+                      (FileTime) Files.getAttribute(Path.of(s.getAbsolutePath()), "creationTime");
+                  // Get the latest file creation time for completion date
```
callunity, line 250: Could similarly use a `.max` function on the stream, and then set the DTO's completionDate

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,335 @@
...
+                cellDirectory ->
+                    cellDirectory.isDirectory()
+                        && REVIO_CELL_DIRECTORY.test(cellDirectory.getName())))
+        .flatMap(
+            cellDirectory ->
+                Arrays.stream(
+                    cellDirectory.listFiles(
+                        metadataDirectory ->
+                            metadataDirectory.isDirectory()
+                                && metadataDirectory.getName().equals("metadata"))));
```
callunity, line 333: could you use a `filter` call here?

callunity:
Are you able to add tests to this? It would be great to see it working over some real data. Dillan would be the best person to advise on how to modify real instrument output to be appropriate for uploading to GitHub.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/PacBioNotificationDto.java
```diff
@@ -9,6 +10,8 @@ public class PacBioNotificationDto extends NotificationDto {
 
   private Map<String, String> poolNames;
 
+  private List<Map<String, String>> positionList;
```
djcooke, line 13: Create a proper model class for this instead of using a map to write a set of undefined fields

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/PacBioNotificationDto.java
```diff
@@ -10,7 +10,10 @@ public class PacBioNotificationDto extends NotificationDto {
 
   private Map<String, String> poolNames;
 
-  private List<Map<String, String>> positionList;
+  public record SMRTCellPosition(
+      String resultFolder, String position, String containerSerialNumber, String poolName) {}
```
callunity, line 14: Will the `containerSerialNumber` be different across different `position`s?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/PacBioNotificationDto.java
```diff
@@ -45,8 +52,19 @@ public void setPoolNames(Map<String, String> poolNames) {
...
 
+  public void setPositionList(List<Map<String, String>> positionList) {
+    this.positionList = positionList;
+  }
+
   @Override
   public String toString() {
-    return super.toString() + ", PacBioNotificationDto [poolNames = " + poolNames + "]";
+    return super.toString()
+        + ", PacBioNotificationDto [poolNames = "
```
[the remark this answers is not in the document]
callunity: Did you push this change to GitHub? It's still looking like multiple arrays on the PR diff.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,346 @@
...
+    // Presence of Transfer_Test_*.txt indicates the run has started, get the creation time
+    Optional<File> mostRecentTransferTestFile =
+        getMetadataDirectory(runDirectory)
+            .flatMap(
+                metadataDirectory ->
+                    Stream.of(
+                        Objects.requireNonNull(
+                            metadataDirectory.listFiles(
+                                file -> TRANSFER_TEST.test(file.getName())))))
+            .min(Comparator.comparing(RevioPacBioProcessor::getFileCreationTime));
```
callunity, line 198: Why `min`? With the result being the most recent transfer test file, I expected it might be `max`

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,346 @@
...
+
+    // Set completion time based on most recently created .Transfer_Test file
+    dto.setStartDate(
+        getFileCreationTime(Objects.requireNonNull(mostRecentTransferTestFile.orElse(null))));
+
+    // Set completion time based on most recently created .Transfer_Test file
+    dto.setStartDate(
+        getFileCreationTime(Objects.requireNonNull(mostRecentTransferTestFile.orElse(null))));
```
callunity, line 206: It looks like these two calls are identical, and do a different thing than the comment indicates.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
callunity: That all makes sense, thanks!

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
```diff
@@ -120,8 +129,18 @@ public static ObjectMapper createObjectMapper() {
 
   public static Optional<Document> parseXml(File file) {
     try {
-      return Optional.of(DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(file));
-    } catch (SAXException | ParserConfigurationException | IOException e) {
+      DocumentBuilder builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+      // Automatically detect BOMs and remove them from input stream
+      // BOM characters can interfere with text processing if not properly removed
+      try (BOMInputStream bomInputStream = new BOMInputStream(new FileInputStream(file));
+          Reader reader = new InputStreamReader(bomInputStream, StandardCharsets.UTF_8)) {
```
callunity, line 136: I'm not sure it's ideal to force all files to be read in UTF-8 by default. There could be some well-formed documents that are not in UTF-8 and could still be read by the default parser (like if the document were actually in UTF-16 as it purports to be - there's no reason they couldn't fix that bug by setting it to be encoded in UTF-16).

I'm thinking that you could separate this out into two functions - `naivelyParseXml` which wraps the original `Optional.of(DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(file));` in a `try` and returns the empty optional if it `catch`-es an error. The second function `utf8ParseXml` would do the forcing you've laid out here, and would return the empty optional if it `catch`-es an error. The larger `parseXml` function could call `naivelyParseXml` then, if empty optional, return the results of `utf8ParseXml`. This would allow a retry and still end up with empty optional if things go wrong.

@djcooke I'd really like your input here.

callunity: But if there's an actual error in the doc, forcing UTF-8 parsing won't necessarily succeed and it'll throw the same error again (SAXException)

djcooke: Add another try/catch there. Any exception when trying to force UTF-8 parsing should result in error log and returning empty too

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
```diff
@@ -127,25 +126,32 @@ public static ObjectMapper createObjectMapper() {
     return mapper;
   }
 
+  /** Attempt to parse naively, and force parsing using UTF-8 if that doesn't work */
   public static Optional<Document> parseXml(File file) {
     try {
-      DocumentBuilder builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+      return Optional.of(DocumentBuilderFactory.newInstance().newDocumentBuilder().parse(file));
+    } catch (ParserConfigurationException | IOException | SAXException e) {
+      log.error("Not really a UTF-16 parsing exception, forcing UTF-8 parsing...", e);
```
djcooke, line 134: - Which exception is actually thrown when this happens? Only that exception should be handled in this way. Fail fast for the others by catching in a separate block (which you already have below catching `Exception`).
- Change to warning level. It's not an error at this point since we're attempting to handle it.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
```diff
@@ -127,25 +126,32 @@ public static ObjectMapper createObjectMapper() {
...
+      DocumentBuilder builder;
+      try {
+        builder = DocumentBuilderFactory.newInstance().newDocumentBuilder();
+      } catch (ParserConfigurationException e2) {
+        throw new RuntimeException(e2);
```
djcooke, line 139: I'm guessing it's not `ParserConfigurationException` that's thrown when the specified encoding is wrong, so you can probably move it to the fail-fast catch block and remove this inner try/catch.

Either way, you should only be throwing a `RuntimeException` here if you expect this to never happen, there should be an exception message that explains this, and you shouldn't be catching that exception below.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
```diff
@@ -127,25 +126,32 @@ public static ObjectMapper createObjectMapper() {
...
-      } catch (FileNotFoundException | SAXException e) {
-        log.error("Failed to parse XML", e);
+      } catch (SAXException | IOException e3) {
+        log.error("Failed to parse XML", e3);
         return Optional.empty();
       }
-
-    } catch (ParserConfigurationException | IOException e) {
-      log.error("Failed to parse XML", e);
+    } catch (Exception e4) {
```
djcooke, line 150: Catch the specific exceptions that can happen here

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
```diff
@@ -127,25 +126,32 @@ public static ObjectMapper createObjectMapper() {
...
+      }
       // Automatically detect BOMs and remove them from input stream
       // BOM characters can interfere with text processing if not properly removed
       try (BOMInputStream bomInputStream = new BOMInputStream(new FileInputStream(file));
           Reader reader = new InputStreamReader(bomInputStream, StandardCharsets.UTF_8)) {
         return Optional.of(builder.parse(new InputSource(reader)));
-      } catch (FileNotFoundException | SAXException e) {
-        log.error("Failed to parse XML", e);
+      } catch (SAXException | IOException e3) {
+        log.error("Failed to parse XML", e3);
```
djcooke, line 147:
```suggestion
        log.error("Failed to parse XML after forcing UTF-8 encoding", e3);
```

Having a different message will help with debugging

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,413 @@
...
+              .toList();
+
+      // Need to take the list of file paths and get the earliest creation time
+      Optional<Path> earliestCreatedFilePath =
+          transferTestFiles
+              .stream()
+              .min(Comparator.comparing(filePath -> getFileCreationTime(filePath.toFile())));
+      dto.setStartDate(
+          getFileCreationTime(
+              Objects.requireNonNull(earliestCreatedFilePath.orElse(null)).toFile()));
```
callunity, line 222: Are transferTest files guaranteed to be present when runscanner first detects a Revio run? (Otherwise you risk an unhandled NPE here)

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,413 @@
...
+        .forEach(metadata -> processMetadata(metadata.get(), dto, tz));
+
+    // Check for .transferdone and Transfer_Test in all SMRT Cells
+    // to consider the run complete
+    if (isRunComplete(runDirectory, smrtCellCount)) {
+      dto.setHealthType(HealthType.COMPLETED);
+
+      /*
+       We have all the expected files. Initially use file creation time, but if
+       there is a timestamp from pbereports.log, use that instead
```
callunity, line 244: Why did you do it in this order instead of checking for pbereports.log first, then file creation time as a backup?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
djcooke: It's a file. There's never any guarantee 😆

callunity (approved):

scanner/pom.xml
```diff
@@ -109,6 +109,11 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+      <version>2.16.1</version>
```
djcooke, line 115: Versions should be managed in the root pom.xml in `<dependencyManagement>`

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,413 @@
...
+    };
+  }
+
+  /**
+   * Extract a number from the metadata file and put the result into the DTO.
+   *
+   * @param expression the XPath expression yielding the number
+   * @param setter the writer for the number
+   */
+  private static RevioPacBioProcessor.ProcessMetadata processNumber(
```
djcooke, line 102: This isn't used for anything

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,413 @@
...
+      String containerSerialNumber = (String) expr[1].evaluate(document, XPathConstants.STRING);
+      String poolName = (String) expr[2].evaluate(document, XPathConstants.STRING);
+      String movieLength = (String) expr[3].evaluate(document, XPathConstants.STRING);
+
+      // From DefaultPacBio, checking to make sure string is valid
+      if (isStringBlankOrNull(position)
+          || isStringBlankOrNull(containerSerialNumber)
+          || isStringBlankOrNull(poolName)
+          || isStringBlankOrNull(movieLength)) {
+        return;
```
djcooke, line 134: If you can't find one piece of information, you don't include any of the other pieces?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
callunity: > Removed `Objects#requireNonNull` from all places

Did this commit get pushed?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -247,39 +207,40 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
-                                  metadataDirectory.listFiles(
-                                      file -> TRANSFER_DONE.test(file.getName())))))
-                  .max(Comparator.comparing(RevioPacBioProcessor::getFileCreationTime));
+                              metadataDirectory.listFiles(
+                                  file -> TRANSFER_DONE.test(file.getName()))))
+                  .map(RevioPacBioProcessor::getFileCreationTime)
+                  .max(Comparator.naturalOrder());
           // Set completion time based on most recently created .transferdone file
-          dto.setCompletionDate(getFileCreationTime(mostRecentDoneFile.orElse(null)));
+          dto.setCompletionDate(mostRecentDoneFile.get());
```
callunity, line 237: Should probably rename this variable to a time-related name since it's no longer referencing a file

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -410,4 +368,24 @@ private boolean isRunComplete(File runDirectory, int smrtCellCount) {
...
+   */
+  private void getStartTimeFromTransferTests(File runDirectory, PacBioNotificationDto dto) {
+    try (Stream<Path> stream = Files.walk(runDirectory.toPath())) {
+      // Need to take the list of file paths and get the earliest creation time
+      Optional<Path> earliestCreatedFilePath =
+          stream
+              .filter(Files::isRegularFile)
+              .filter(file -> TRANSFER_TEST.test(String.valueOf(file.getFileName())))
+              .min(Comparator.comparing(filePath -> getFileCreationTime(filePath.toFile())));
+      dto.setStartDate(getFileCreationTime(earliestCreatedFilePath.orElse(null).toFile()));
```
callunity, line 386: It looks like you're calling `getFileCreationTime` twice here; can you think of a way to simplify this?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,399 @@
...
+  private static Instant getLogCompletionTime(File file) {
+    try {
+      Scanner myReader = new Scanner(file);
+      Pattern pattern =
+          Pattern.compile("^\\[INFO] (.*) \\[.*] exiting with return code \\d+" + " .*");
+      while (myReader.hasNextLine()) {
+        Matcher matcher = pattern.matcher(myReader.nextLine());
+        if (matcher.matches()) {
+          String stringDate = matcher.group(1);
+          String datePattern = "yyyy-MM-dd HH:mm:ss[.SSS][,SSS][Z][X]";
```
djcooke, line 323: - All the examples you've seen have had "Z" for the time offset, right? You can put `XX` for the time offset instead of `[Z][X]` if so - I think that's a better assumption than either `X` or `Z`, and I think both definitely shouldn't be there, and it shouldn't be optional unless you've seen a case where it's missing
- Have you seen both "." and "," to separate the milliseconds? If no, only include the one you've seen, and don't make it optional unless you've seen it missing. If you've seen both, I guess just make the separators optional instead of grouping the `SSS` in twice (i.e. `[,][.]SSS`). I don't think you can do an "or" with date formats unfortunately

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,399 @@
...
+          DateTimeFormatter dateTimeFormatter = DateTimeFormatter.ofPattern(datePattern);
+          return LocalDateTime.parse(stringDate, dateTimeFormatter)
+              .atOffset(ZoneOffset.UTC)
+              .toInstant();
+        }
+      }
+    } catch (FileNotFoundException e) {
+      throw new RuntimeException(e);
```
djcooke, line 331: Throw `IllegalArgumentException` instead with a message saying the file is expected to exist. This should never get thrown currently, but that could change in the future. Look at it from the context of this method itself. There's nothing here to ensure the file exists, which means it's up to the caller.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,399 @@
...
+              .filter(file -> TRANSFER_TEST.test(String.valueOf(file.getFileName())))
+              .toList();
+
+      List<Path> transferDoneFiles =
+          testDoneStream
+              .filter(Files::isRegularFile)
+              .filter(file -> TRANSFER_DONE.test(String.valueOf(file.getFileName())))
+              .toList();
+
+      return transferDoneFiles.size() == smrtCellCount && transferTestFiles.size() == smrtCellCount;
```
djcooke, line 356: Can use `Stream#count` instead of `toList` since you don't need the lists

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,399 @@
...
+   * @return Instant point in time
+   */
+  private static Instant getFileCreationTime(File file) {
+    try {
+      FileTime fileTime =
+          (FileTime) Files.getAttribute(Path.of(file.getAbsolutePath()), "creationTime");
+      return fileTime.toInstant();
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
djcooke, line 305: Change this method to throw `IOException` instead of catching it. The `process` method is already expected to throw it, so it's fine to let this bubble up

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,399 @@
...
+      List<Path> transferDoneFiles =
+          testDoneStream
+              .filter(Files::isRegularFile)
+              .filter(file -> TRANSFER_DONE.test(String.valueOf(file.getFileName())))
+              .toList();
+
+      return transferDoneFiles.size() == smrtCellCount && transferTestFiles.size() == smrtCellCount;
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
djcooke, line 359: Make method throw `IOException` instead of catching

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,399 @@
...
+    Optional<Instant> earliestTransferTestCreationTime =
+        streamSmrtCellSubdirectories(runDirectory, "metadata")
+            .flatMap(
+                metadataDirectory ->
+                    Stream.of(
+                        metadataDirectory.listFiles(file -> TRANSFER_TEST.test(file.getName()))))
+            .map(RevioPacBioProcessor::getFileCreationTime)
+            .min(Comparator.naturalOrder());
+    // Set run start time based on most recently created .Transfer_Test file
+    dto.setStartDate(earliestTransferTestCreationTime.get());
```
djcooke, line 378: Will throw `NoSuchElementException` if the files are missing. That's probably fine for now, but you should document it here

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -0,0 +1,399 @@
...
+    Optional<Instant> latestTransferDoneCreationTime =
+        streamSmrtCellSubdirectories(runDirectory, "metadata")
+            .flatMap(
+                metadataDirectory ->
+                    Stream.of(
+                        metadataDirectory.listFiles(file -> TRANSFER_DONE.test(file.getName()))))
+            .map(RevioPacBioProcessor::getFileCreationTime)
+            .max(Comparator.naturalOrder());
+    // Set completion time based on most recently created .transferdone file
+    dto.setCompletionDate(latestTransferDoneCreationTime.get());
```
djcooke, line 397: Document possible `NoSuchElementException`

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -320,15 +307,15 @@ private static Instant getLogCompletionTime(File file) {
         Matcher matcher = pattern.matcher(myReader.nextLine());
         if (matcher.matches()) {
           String stringDate = matcher.group(1);
-          String datePattern = "yyyy-MM-dd HH:mm:ss[.SSS][,SSS][Z][X]";
+          String datePattern = "yyyy-MM-dd HH:mm:ss[,SSS][XX]";
```
djcooke, line 310: Why the optional parts? Have you seen examples where these are missing?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -364,6 +351,7 @@ private boolean isRunComplete(File runDirectory, int smrtCellCount) {
    *
    * @param runDirectory which we are currently processing
    * @param dto PacBioNotificationDto which we will modify
+   *     <p>Throws NoSuchElementException if the .Transfer_Test files are missing
```
djcooke, line 354: Use `@throws` to include this properly in JavaDoc

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -383,6 +371,7 @@ private void setStartTimeFromTransferTest(File runDirectory, PacBioNotificationD
    *
    * @param runDirectory which we are currently processing
    * @param dto PacBioNotificationDto which we will modify
+   *     <p>Throws NoSuchElementException if the .Transferdone files are missing
```
djcooke, line 374: `@throws`

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
djcooke: Missed this one

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -350,8 +344,8 @@ private boolean isRunComplete(File runDirectory, int smrtCellCount) throws IOExc
    * Check for presence of Transfer_Test file and grab creation time
    *
    * @param runDirectory which we are currently processing
-   * @param dto PacBioNotificationDto which we will modify
-   *     <p>Throws NoSuchElementException if the .Transfer_Test files are missing
+   * @param dto PacBioNotificationDto which we will modify @Throws NoSuchElementException if the
```
djcooke, line 347: `@throws` needs to be lowercase and on its own line. If you hover over the method name in your IDE, it should pop up the JavaDoc, so you can see it's formatted correctly. Same on L375

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
djcooke: The easiest good solution would be to do this outside of a stream

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -341,58 +331,54 @@ private boolean isRunComplete(File runDirectory, int smrtCellCount) throws IOExc
...
 
   /**
-   * Check for presence of Transfer_Test file and grab creation time
+   * Filter runDirectory for Transfer_Test files and grab earliest creation time
    *
    * @param runDirectory which we are currently processing
-   * @param dto PacBioNotificationDto which we will modify @Throws NoSuchElementException if the
-   *     .Transfer_Test files are missing
+   * @return earliest file creation time
+   * @throws NoSuchElementException if the .Transfer_Test files are missing
```
djcooke, line 338: This is no longer thrown, and the `@return` should include "or null if no Transfer_Test files are found"

Same for `completionTimeFromTransferDone`.

I'm not sure whether failing to find a start time should cause an error... Can leave as is for now

djcooke (approved):
Approved once JavaDocs are fixed

PR #177, opened 2025-03-10
GLT-4353 fix runs not displaying

Christopher-Ho:
JIRA Ticket: https://jira.oicr.on.ca/browse/GLT-4353

- [x] Updates release notes

### DIFF
--- a/scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessorTest.java
+++ b/scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessorTest.java
@@ -32,10 +32,9 @@ public class RevioPacBioProcessorTest<T> extends AbstractProcessorTest<PacBioNot
       // Assert that a start time (any) was detected
       Assert.assertNotNull(result.getStartDate());
 
-      // Need to modify reference.json because for runs that have just started, the only
-      // information they have are the run name and start time using file creation. For tests, we
-      // cannot use file creation time as these aren't stored in git.
-      // Set start date to hard-coded value matching reference.json
+      // Need to modify result to match reference.json because for runs that have just started,
+      // the only information available are run name and start time using file creation. For
+      // tests, we cannot use file creation time as these aren't stored in git
       result.setStartDate(
           LocalDateTime.parse(
                   "2025-03-10 20:11:46,494376012Z",


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessor.java
```diff
@@ -180,11 +185,25 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+                                .matches("^([A-Za-z0-9]+(_[A-Za-z0-9]+){3})\\.metadata\\.xml$")))
         .map(RunProcessor::parseXml)
         .filter(Optional::isPresent)
         .forEach(metadata -> processMetadata(metadata.get(), dto, tz));
 
+    // Check if we grabbed a run alias from metadata.xml
+    if (dto.getRunAlias() != null) {
+      // Not valid, if it doesn't match run directory name
+      if (!runDirectory.getName().equals(dto.getRunAlias())) {
+        dto.setRunAlias(null);
```
djcooke, line 200: This is an error that makes the run invalid, so throw a runtime exception that states that. Once we know a run is invalid, we should stop wasting time processing it.

Would also be good to leave comment about why this is necessary since it's not obvious - when the run is first started, we can only get the name from the directory, so we need to make sure the same run doesn't show up under a different name later when the metadata files are available.

scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessorTest.java
```diff
@@ -0,0 +1,44 @@
...
+  protected NotificationDto process(File directory) throws IOException {
+    afterProcessing = instance.process(directory, TimeZone.getTimeZone("America/Toronto"));
+    return afterProcessing;
+    // return instance.process(directory, TimeZone.getTimeZone("America/Toronto"));
+  }
+
+  @Override
+  public void beforeComparison(PacBioNotificationDto reference, PacBioNotificationDto result) {
+    // Check if run name matches new run directory
+    if (afterProcessing.getRunAlias().equals(result.getRunAlias())) {
```
callunity, line 31: What should happen if it doesn't?

scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessorTest.java
```diff
@@ -0,0 +1,44 @@
...
+      // Assert that a start time (any) was detected
+      Assert.assertNotNull(afterProcessing.getStartDate());
+
+      // Set the start date to match the reference json
+      afterProcessing.setStartDate(reference.getStartDate());
```
callunity, line 36: Please add some explanation as to why you are modifying the data mid-test.

scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/AbstractProcessorTest.java
```diff
@@ -21,14 +21,18 @@
...
+  private final Class<T> clazz;
 
-  public AbstractProcessorTest(Class<? extends NotificationDto> clazz) {
+  public AbstractProcessorTest(Class<T> clazz) {
     super();
     this.clazz = clazz;
   }
 
+  public void beforeComparison(T reference, T result) {
+    // Not used by default
```
djcooke, line 33: It does nothing by default, but it is used - `checkDirectory` always calls it

scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessorTest.java
```diff
@@ -0,0 +1,47 @@
...
+  protected NotificationDto process(File directory) throws IOException {
+    afterProcessing = instance.process(directory, TimeZone.getTimeZone("America/Toronto"));
+    return afterProcessing;
+    // return instance.process(directory, TimeZone.getTimeZone("America/Toronto"));
+  }
+
+  @Override
+  public void beforeComparison(PacBioNotificationDto reference, PacBioNotificationDto result) {
+    // Check if reference run name matches new run directory name
+    if (afterProcessing.getRunAlias().equals(reference.getRunAlias())) {
```
djcooke, line 31: 1. This should only be for one specific run, so you should have the name of that run hard-coded here. As is, you're changing the start time of every Revio test run
2. There's no need for`afterProcessing` - it's the same object as `result`

scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/RevioPacBioProcessorTest.java
```diff
@@ -0,0 +1,47 @@
...
+      // Assert that a start time (any) was detected
+      Assert.assertNotNull(afterProcessing.getStartDate());
+
+      // Need to modify reference.json because for runs that have just started, the only
+      // information they have are the run name and start time using file creation. For tests, we
+      // cannot use file creation time as these aren't stored in git.
+      // Set the start date to match the reference json
+      afterProcessing.setStartDate(reference.getStartDate());
```
djcooke, line 39: Also hard-code the date you're setting to make sure it's doing what you think it is.

PR #191, opened 2025-12-12
Scan Illumina run directories in reverse order (which is approximately reverse chronological order)

mlaszloffy:
I don't think we want to use directory last modified, as we perform run directory cleaning that will cause older run directories to be scanned ahead of newer run directories.

Change the workpool/runscanner job executor to a FixedThreadPool with a default LinkedBlockingQueue to ensure jobs get scanned in a FIFO basis (so the job execution order is in the same sort order of the Illumina run directories).

JIRA Ticket:

- [x] Updates release notes


mlaszloffy:
Round robin scanning will be looked at in https://jira.oicr.on.ca/browse/GP-5578

avarsava (approved):

grapigeau (approved):

djcooke (approved):

PR #201, opened 2026-05-08
Gp 5410.minimal ultima support

grapigeau:
JIRA Ticket: https://jira.oicr.on.ca/browse/GP-5410

- [x] Updates release notes

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,7 @@
 Changes:
 
  * Update read the doc links
+ * Added initial support for Ultima UG 100 run information
  
 # 2.4.0
 
--- a/docs/appendices.md
+++ b/docs/appendices.md
@@ -16,6 +16,7 @@
             <li>ILLUMINA</li>
             <li>PACBIO</li>
             <li>OXFORDNANOPORE</li>
+            <li>ULTIMA</li>
         </ul></td>
     </tr>
     <tr>
@@ -37,6 +38,11 @@
             <li>"minion"</li>
             <li>"testing"</li>
         </ul></td>
+        <td>Ultima:
+        <ul>
+            <li>"default"</li>
+            <li>"testing"</li>
+        </ul>
     </tr>
     <tr>
         <td>timeZone</td>
@@ -81,6 +87,17 @@
                 <td>(no additional parameters)</td>
                 <td></td>
             </tr>
+            <tr colspan="2">
+                <th>ULTIMA</th>
+            </tr>
+            <tr>
+                <td>nexusApiAddress (required)</td>
+                <td>The base URL of the Nexus API (Ex. https://example-nexus.com)</td>
+            </tr>
+            <tr>
+                <td>nexusApiTokenFile (required)</td>
+                <td>The location of a file containing a Nexus Auth Token</td>
+            </tr>
         </table></td>
     </tr>
 </table>
@@ -95,9 +112,11 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <th>PacBio</th>
         <th>Illumina</th>
         <th>Oxford Nanopore</th>
+        <th>Ultima</th>
     </tr>
     <tr>
         <td>platform</td>
+        <td>✓</td>
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
@@ -107,15 +126,18 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
+        <td>✓</td>
     </tr>
     <tr>
         <td>sequencerFolderPath</td>
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
+        <td></td>
     </tr>
     <tr>
         <td>sequencerName</td>
+        <td>✓</td>
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
@@ -125,9 +147,11 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
+        <td>✓</td>
     </tr>
     <tr>
         <td>containerSerialNumber</td>
+        <td>✓</td>
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
@@ -137,9 +161,11 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
+        <td>✓</td>
     </tr>
     <tr>
         <td>laneCount</td>
+        <td>✓</td>
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
@@ -149,9 +175,11 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
+        <td>✓</td>
     </tr>
     <tr>
         <td>startDate</td>
+        <td>✓</td>
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
@@ -161,9 +189,11 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
+        <td>✓</td>
     </tr>
     <tr>
         <td>pairedEndRun</td>
+        <td>✓</td>
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
@@ -173,9 +203,11 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
+        <td>✓</td>
     </tr>
     <tr>
         <td>metrics</td>
+        <td>✓</td>
         <td>✓</td>
         <td>✓</td>
         <td>✓</td>
@@ -183,6 +215,7 @@ esp bc they might vary in meaning b/w PromION and MinION -->
     <tr>
         <td>poolNames</td>
         <td>✓</td>
+        <td></td>
         <td></td>
         <td></td>
     </tr>
@@ -191,11 +224,13 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td></td>
         <td>✓</td>
         <td></td>
+        <td></td>
     </tr>
     <tr>
         <td>bclCount</td>
         <td></td>
         <td>✓</td>
+        <td></td>
         <td></td>
     </tr>
     <tr>
@@ -203,11 +238,13 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td></td>
         <td>✓</td>
         <td></td>
+        <td></td>
     </tr>
     <tr>
         <td>chemistry</td>
         <td></td>
         <td>✓</td>
+        <td></td>
         <td></td>
     </tr>
     <tr>
@@ -215,11 +252,13 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td></td>
         <td>✓</td>
         <td></td>
+        <td></td>
     </tr>
     <tr>
         <td>indexLengths</td>
         <td></td>
         <td>✓</td>
+        <td></td>
         <td></td>
     </tr>
     <tr>
@@ -227,11 +266,13 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td></td>
         <td>✓</td>
         <td></td>
+        <td></td>
     </tr>
     <tr>
         <td>numReads</td>
         <td></td>
         <td>✓</td>
+        <td></td>
         <td></td>
     </tr>
     <tr>
@@ -239,17 +280,20 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td></td>
         <td>✓</td>
         <td></td>
+        <td></td>
     </tr>
     <tr>
         <td>readLength</td>
         <td></td>
         <td>✓</td>
         <td></td>
+        <td>✓</td>
     </tr>
     <tr>
         <td>scoreCycle</td>
         <td></td>
         <td>✓</td>
+        <td></td>
         <td></td>
     </tr>
     <tr>
@@ -257,15 +301,60 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td></td>
         <td>✓</td>
         <td></td>
+        <td></td>
     </tr>
     <tr>
         <td>runType</td>
         <td></td>
         <td></td>
         <td>✓</td>
+        <td></td>
     </tr>
     <tr>
         <td>protocolVersion</td>
+        <td></td>
+        <td></td>
+        <td>✓</td>
+        <td></td>
+    </tr>
+    <tr>
+        <td>sequencingStatus</td>
+        <td></td>
+        <td></td>
+        <td></td>
+        <td>✓</td>
+    </tr>
+    <tr>
+        <td>analysisStatus</td>
+        <td></td>
+        <td></td>
+        <td></td>
+        <td>✓</td>
+    </tr>
+    <tr>
+        <td>uploadStatus</td>
+        <td></td>
+        <td></td>
+        <td></td>
+        <td>✓</td>
+    </tr>
+    <tr>
+        <td>expectedFlows</td>
+        <td></td>
+        <td></td>
+        <td></td>
+        <td>✓</td>
+    </tr>
+    <tr>
+        <td>completedFlows</td>
+        <td></td>
+        <td></td>
+        <td></td>
+        <td>✓</td>
+    </tr>
+    <tr>
+        <td>waferShelf</td>
+        <td></td>
         <td></td>
         <td></td>
         <td>✓</td>
@@ -275,10 +364,12 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
         <td></td>
         <td></td>
+        <td></td>
     </tr>
     <tr>
         <td>pipelineRuns</td>
         <td>✓</td>
+        <td></td>
         <td></td>
         <td></td>
     </tr>
@@ -323,6 +414,13 @@ unsupported.
 * **NovaSeq X Plus**
 * iSeq
 * MiniSeq
+
+#### Ultima - "default" processor
+* *UG 100*
+
+**Notes for Ultima Sequencers**
+
+* Current support is in development and subject to change
 
 #### Oxford Nanopore - "promethion" processor
 
--- a/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
+++ b/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
@@ -1,0 +1,104 @@
+package ca.on.oicr.gsi.runscanner.dto;
+
+import ca.on.oicr.gsi.runscanner.dto.type.Platform;
+import ca.on.oicr.gsi.runscanner.dto.type.UltimaProcessStatus;
+import java.util.Objects;
+
+public class UltimaNotificationDto extends NotificationDto {
+
+  private int completedFlows;
+  private int expectedFlows;
+  private double readLength;
+  private int waferShelf;
+  private UltimaProcessStatus sequencingStatus;
+  private UltimaProcessStatus analysisStatus;
+  private UltimaProcessStatus uploadStatus;
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) return true;
+    if (!super.equals(obj)) return false;
+    if (getClass() != obj.getClass()) return false;
+    UltimaNotificationDto other = (UltimaNotificationDto) obj;
+
+    return Objects.equals(expectedFlows, other.expectedFlows)
+        && Objects.equals(waferShelf, other.waferShelf);
+  }
+
+  @Override
+  public Platform getPlatformType() {
+    return Platform.ULTIMA;
+  }
+
+  public int getCompletedFlows() {
+    return completedFlows;
+  }
+
+  public void setCompletedFlows(int completedFlows) {
+    this.completedFlows = completedFlows;
+  }
+
+  public int getExpectedFlows() {
+    return expectedFlows;
+  }
+
+  public void setExpectedFlows(int expectedFlows) {
+    this.expectedFlows = expectedFlows;
+  }
+
+  public double getReadLength() {
+    return readLength;
+  }
+
+  public void setReadLength(double readLength) {
+    this.readLength = readLength;
+  }
+
+  public int getWaferShelf() {
+    return waferShelf;
+  }
+
+  public void setWaferShelf(int waferShelf) {
+    this.waferShelf = waferShelf;
+  }
+
+  public UltimaProcessStatus getSequencingStatus() {
+    return sequencingStatus;
+  }
+
+  public void setSequencingStatus(UltimaProcessStatus sequencingStatus) {
+    this.sequencingStatus = sequencingStatus;
+  }
+
+  public UltimaProcessStatus getAnalysisStatus() {
+    return analysisStatus;
+  }
+
+  public void setAnalysisStatus(UltimaProcessStatus analysisStatus) {
+    this.analysisStatus = analysisStatus;
+  }
+
+  public UltimaProcessStatus getUploadStatus() {
+    return uploadStatus;
+  }
+
+  public void setUploadStatus(UltimaProcessStatus uploadStatus) {
+    this.uploadStatus = uploadStatus;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(super.hashCode(), expectedFlows, waferShelf);
+  }
+
+  @Override
+  public String toString() {
+    return super.toString()
+        + ", UltimaNotificationDto ["
+        + ", expectedFlows="
+        + expectedFlows
+        + ", wafershelf="
+        + waferShelf
+        + "]";
+  }
+}
--- a/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/type/Platform.java
+++ b/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/type/Platform.java
@@ -27,5 +27,6 @@ package ca.on.oicr.gsi.runscanner.dto.type;
 public enum Platform {
   ILLUMINA, //
   PACBIO, //
-  OXFORDNANOPORE
+  OXFORDNANOPORE, //
+  ULTIMA, //
 }
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Configuration.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Configuration.java
@@ -1,7 +1,9 @@
 package ca.on.oicr.gsi.runscanner.scanner;
 
 import ca.on.oicr.gsi.Pair;
+import ca.on.oicr.gsi.runscanner.scanner.processor.PathType;
 import ca.on.oicr.gsi.runscanner.scanner.processor.RunProcessor;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
@@ -10,13 +12,16 @@ import java.util.stream.Stream;
 
 public class Configuration {
 
+  private List<File> ignoreSubdirectories;
   private File path;
-
   private RunProcessor processor;
+  private ObjectNode parameters;
 
   private TimeZone timeZone;
 
-  private List<File> ignoreSubdirectories;
+  public List<File> getIgnoreSubdirectories() {
+    return ignoreSubdirectories;
+  }
 
   public File getPath() {
     return path;
@@ -24,6 +29,10 @@ public class Configuration {
 
   public RunProcessor getProcessor() {
     return processor;
+  }
+
+  public ObjectNode getParameters() {
+    return parameters;
   }
 
   public Stream<Pair<File, Configuration>> getRuns() {
@@ -34,11 +43,13 @@ public class Configuration {
     return timeZone;
   }
 
-  public List<File> getIgnoreSubdirectories() {
-    return ignoreSubdirectories;
-  }
-
   public boolean isValid() {
+
+    // If the path is VIRTUAL, we only care that a path string and timezone are provided
+    if (processor != null && processor.getPathType() == PathType.VIRTUAL) {
+      return path != null && timeZone != null && processor.validateParameters(parameters);
+    }
+
     return path != null
         && path.isDirectory()
         && path.canRead()
@@ -57,20 +68,12 @@ public class Configuration {
       if (!path.canExecute()) summary += "Path cannot be executed! ";
     }
     if (processor == null) summary += "Processor is null! ";
-    if (timeZone == null) summary += "TimeZone is null!";
+    if (timeZone == null) summary += "TimeZone is null! ";
+    if (processor != null && processor.getPathType() == PathType.VIRTUAL) {
+      if (!processor.validateParameters(parameters))
+        summary += "Parameters not configured correctly! ";
+    }
     return summary;
-  }
-
-  public void setPath(File path) {
-    this.path = path;
-  }
-
-  public void setProcessor(RunProcessor processor) {
-    this.processor = processor;
-  }
-
-  public void setTimeZone(TimeZone timeZone) {
-    this.timeZone = timeZone;
   }
 
   public void setIgnoreSubdirectories(List<File> ignoreSubdirectories) {
@@ -79,5 +82,21 @@ public class Configuration {
     } else {
       this.ignoreSubdirectories = ignoreSubdirectories;
     }
+  }
+
+  public void setPath(File path) {
+    this.path = path;
+  }
+
+  public void setParameters(ObjectNode parameters) {
+    this.parameters = parameters;
+  }
+
+  public void setProcessor(RunProcessor processor) {
+    this.processor = processor;
+  }
+
+  public void setTimeZone(TimeZone timeZone) {
+    this.timeZone = timeZone;
   }
 }
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
@@ -1,0 +1,280 @@
+package ca.on.oicr.gsi.runscanner.scanner.processor;
+
+import ca.on.oicr.gsi.runscanner.dto.NotificationDto;
+import ca.on.oicr.gsi.runscanner.dto.UltimaNotificationDto;
+import ca.on.oicr.gsi.runscanner.dto.type.HealthType;
+import ca.on.oicr.gsi.runscanner.dto.type.UltimaProcessStatus;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.io.File;
+import java.io.IOException;
+import java.math.BigDecimal;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultUltima extends RunProcessor {
+
+  // Cache to store API results during a single scan cycle
+  protected final Map<String, JsonNode> runCache = new ConcurrentHashMap<>();
+
+  DateTimeFormatter NEXUS_DATE_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+
+  private static final Logger log = LoggerFactory.getLogger(DefaultUltima.class);
+  private final UltimaApiClient apiClient;
+
+  public static DefaultUltima create(Builder builder, ObjectNode parameters) {
+    return new DefaultUltima(
+        builder, fetchNexusApiUrl(parameters), fetchNexusApiTokenFile(parameters));
+  }
+
+  protected DefaultUltima(Builder builder, String apiUrl, String tokenPath) {
+    super(builder);
+    this.apiClient = new UltimaApiClient(apiUrl, tokenPath);
+  }
+
+  protected DefaultUltima(Builder builder, UltimaApiClient apiClient) {
+    super(builder);
+    this.apiClient = apiClient;
+  }
+
+  /**
+   * @param parameters ObjectNode
+   * @return String with base URL for the ULTIMA API
+   */
+  private static String fetchNexusApiUrl(ObjectNode parameters) {
+    if (parameters.hasNonNull("nexusApiAddress")) {
+      return parameters.get("nexusApiAddress").asText();
+    } else {
+      log.error("No Nexus API URL configured for Ultima, this config should be invalid");
+      return null;
+    }
+  }
+
+  /**
+   * @param parameters ObjectNode
+   * @return String filename where token is stored
+   */
+  private static String fetchNexusApiTokenFile(ObjectNode parameters) {
+    if (parameters.hasNonNull("nexusApiTokenFile")) {
+      return parameters.get("nexusApiTokenFile").asText();
+    } else {
+      log.error("No Nexus API Token configured for Ultima, this config should be invalid");
+      return null;
+    }
+  }
+
+  /**
+   * Instead of scanning disk, we fetch the list of Run IDs from the API. We return "Virtual" File
+   * objects representing each Run ID.
+   */
+  @Override
+  public Stream<File> getRunsFromRoot(File root) {
+
+    try {
+      List<JsonNode> allRunInfo = apiClient.fetchAllRunSummaries();
+
+      return allRunInfo.stream()
+          .map(
+              node -> {
+                String runId = node.path("runid").asText();
+
+                // Skip if runId is null, empty, or just whitespace
+                if (runId == null || runId.trim().isEmpty()) {
+                  return null;
+                }
+
+                runCache.put(runId, node); // Cache for the process() step
+
+                // TODO eventually will be google bucket path with runID in subfolder name
+                return new File(root, runId);
+              })
+          .filter(Objects::nonNull);
+    } catch (Exception e) {
+      log.error("Failed to fetch run list from Ultima API", e);
+      return Stream.empty();
+    }
+  }
+
+  @Override
+  public NotificationDto process(File runDirectory, TimeZone tz) throws IOException {
+    String runId = runDirectory.getName();
+
+    // Get the run information from the cached API call
+    JsonNode json = runCache.get(runId);
+
+    if (json == null) {
+      throw new IOException("Data for run " + runId + " missing from cache. API sync failure.");
+    }
+
+    UltimaNotificationDto dto = new UltimaNotificationDto();
+    dto.setRunAlias(runId);
+    dto.setSequencerName(json.path("sysid").asText());
+    dto.setSoftware(json.path("SequencingRecipe").asText());
+
+    dto.setCompletedFlows(json.path("completedflownum").asInt());
+    dto.setExpectedFlows(json.path("numflows").asInt());
+    dto.setReadLength(json.path("rl").asDouble());
+    dto.setWaferShelf(json.path("wafershelf").asInt());
+
+    ObjectMapper mapper = createObjectMapper();
+    List<ObjectNode> metrics = new ArrayList<>();
+    ObjectNode chartNode = mapper.createObjectNode();
+    chartNode.put("type", "chart");
+    ArrayNode values = chartNode.putArray("values");
+
+    double numReadPass = json.path("num_reads_pass_filter").asDouble();
+    double numBeads = json.path("numbeads").asDouble();
+    double passFilterPercent = (numBeads > 0) ? (numReadPass / numBeads) * 100 : 0;
+
+    values
+        .addObject()
+        .put("name", "Number of Beads")
+        .put("value", new BigDecimal(numBeads).toPlainString()); // prevent scientific notation
+    values
+        .addObject()
+        .put("name", "Output Reads")
+        .put("value", json.path("pf_output_reads").asText("0"));
+    values
+        .addObject()
+        .put("name", "Pass Filter %")
+        .put("value", String.format("%.2f", passFilterPercent));
+    metrics.add(chartNode);
+    dto.setMetrics(mapper.writeValueAsString(metrics));
+
+    dto.setSequencerPosition(json.path("Chuck").asText());
+
+    String wafer = json.path("wafer").asText();
+    dto.setContainerSerialNumber(wafer);
+    dto.setContainerModel(extractModel(wafer));
+
+    int pctCompleted = json.path("runstatus").asInt(); // Percentage 0-100
+    int succeeded = json.path("isruncompleted").asInt(); // Successful run 0 or 1
+    String errmsg = json.path("errormsg").asText();
+
+    // the first and last digit are vestigial and should be ignored.
+    // the second and third digit are the analysis and upload status respectively.
+    // 0 is not started, 1 is in progress, 2 is complete, and 3+ is error
+    String analAndUpStatus = json.path("analysisstatus").asText();
+
+    UltimaProcessStatus analysisStatus =
+        UltimaProcessStatus.fromCode(Character.getNumericValue(analAndUpStatus.charAt(1)));
+    dto.setAnalysisStatus(analysisStatus);
+    UltimaProcessStatus uploadStatus =
+        UltimaProcessStatus.fromCode(Character.getNumericValue(analAndUpStatus.charAt(2)));
+    dto.setUploadStatus(uploadStatus);
+    UltimaProcessStatus sequencingStatus = sequencingComplete(pctCompleted, succeeded, errmsg);
+    dto.setSequencingStatus(sequencingStatus);
+
+    // Ultima should only be considered complete once upload is complete (crams not available before
+    // that time)
+    HealthType health = translateStatus(sequencingStatus, analysisStatus, uploadStatus);
+    dto.setHealthType(health);
+
+    String zoneString = json.path("timezone").asText();
+    LocalDateTime ldt =
+        LocalDateTime.parse(json.path("startdatetime").asText(), NEXUS_DATE_FORMATTER);
+    Instant startDate = extractTime(ldt, zoneString);
+    dto.setStartDate(startDate);
+    if (health.isDone()) {
+      Duration runtime = Duration.ofMinutes(json.path("runtime").asInt());
+      dto.setCompletionDate(calculateCompletedDate(startDate, runtime));
+    }
+
+    // No lanes on CD wafers (yet)
+    dto.setLaneCount(1);
+    // No paired ends
+    dto.setPairedEndRun(false);
+
+    return dto;
+  }
+
+  private UltimaProcessStatus sequencingComplete(int pctComplete, int succeeded, String errmsg) {
+    if (!errmsg.isEmpty()) {
+      return UltimaProcessStatus.FAILED;
+    } else if (pctComplete == 100) {
+      return (succeeded == 1) ? UltimaProcessStatus.COMPLETE : UltimaProcessStatus.FAILED;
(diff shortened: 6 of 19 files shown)


dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/AnalysisFile.java
```diff
@@ -0,0 +1,95 @@
+package ca.on.oicr.gsi.runscanner.dto;
```
grapigeau: Moved this originally to support Ultima Analysis, that's scoped out of this PR so this looks redundant but it isn't breaking and will be used later

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/type/Platform.java
```diff
@@ -27,5 +27,6 @@
 public enum Platform {
   ILLUMINA, //
   PACBIO, //
-  OXFORDNANOPORE
+  OXFORDNANOPORE,
```
avarsava:
```suggestion
  OXFORDNANOPORE, //
```

I assume these slashes do something w/r/t formatting

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/AnalysisFile.java
avarsava: If it's not relevant, please don't include it. Makes the git history less confusing when looking at the file later.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+public class UltimaNotificationDto extends NotificationDto {
+
+  private int completedFlowNum;
+  private double readLength;
+  private int runFlows;
+  private int wafershelf;
```
avarsava: not `waferShelf`?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
```diff
@@ -0,0 +1,84 @@
...
+
+  public List<PipelineRun> pipelineRuns = new LinkedList<>();
```
avarsava: Unused?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultUltima extends RunProcessor {
+  private static final LatencyHistogram api_fetch_time =
+      new LatencyHistogram(
+          "runscanner_ultima_scan_completness_time", "The time to query the NEXUS API.");
```
avarsava:
```suggestion
          "runscanner_ultima_scan_completeness_time", "The time to query the NEXUS API.");
```

(and in checking why you made the java object snake case, i discovered the illumina metric is also misspelled. we can't change that one now though...

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+  /**
+   * Instead of scanning disk, we fetch the list of Run IDs from the API. We return "Virtual" File
+   * objects representing each Run ID.
+   */
+  @Override
+  public Stream<File> getRunsFromRoot(File root) {
+
+    try (AutoCloseable latency = api_fetch_time.start()) {
+
+      List<JsonNode> metrics = apiClient.fetchAllRunIds();
```
avarsava: calling a list of run ids 'metrics' is a little confusing imo
but it's also a list of jsonnode, so does 'fetchAllRunIds' fetch more than run ids?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+    double numReadPass = json.path("num_reads_pass_filter").asDouble();
+    double numBeads = json.path("numbeads").asDouble();
+
+    ObjectNode runMetrics = metrics.addObject();
+    runMetrics.put("numBeads", numBeads);
+    runMetrics.put("outputReads", Long.parseLong(json.path("pf_output_reads").asText("0")));
+    if (numBeads > 0) {
+      runMetrics.put("passFilterPercent", numReadPass / numBeads);
+    }
+    runMetrics.put("readLength", Double.parseDouble(json.path("rl").asText("0")));
```
avarsava: Duplicating line 134?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+    dto.setContainerSerialNumber(wafer);
+    dto.setContainerModel(extractModel(wafer));
+
+    int status = json.path("runstatus").asInt();
+    int completed = json.path("isruncompleted").asInt();
+    String errmsg = json.path("errormsg").asText();
+    HealthType health = translateStatus(status, completed, errmsg);
+    dto.setHealthType(health);
+
+    // TODO better way for timezone?
```
avarsava: You get passed the user-configured timezone in the method signature, if that helps

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+
+  public String extractModel(String serialNumber) {
+    // Example: FC12.2T0123456789 -> FTC12.2T
+    Pattern pattern = Pattern.compile("^(.{4}\\..{2})");
+    Matcher matcher = pattern.matcher(serialNumber);
+    if (matcher.find()) {
+      // Group 1 contains the extracted text
+      return matcher.group(1);
+    }
+    return serialNumber.substring(0, 7);
```
avarsava: if serialNumber is SOMENONSENSE do you really want to claim the model is SOMENON and carry on? Might want to error in this case instead

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
avarsava: Please update

avarsava: No longer necessarily true, if VIRTUAL always returns true. `isFilePathValid` will return true for a VIRTUAL-type path of "fdslkjfdsafdsa" but the processor probably won't be able to process that.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClientImpl.java
```diff
@@ -0,0 +1,89 @@
...
+        .setConnectTimeout(Duration.ofSeconds(5))
+        .setReadTimeout(Duration.ofSeconds(5))
+        .build();
+  }
+
+  @Override
+  public List<JsonNode> fetchAllRunIds() throws IOException {
+
+    String datetimeRestrictions =
+        "from=01-01-2025&to="
```
avarsava: Is this restriction documented in the manual? See the 'Notes for Oxford Nanopore Sequencers' in the appendix about the 2017 limit

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClientImpl.java
```diff
@@ -0,0 +1,89 @@
...
+      response.getBody().forEach(nodes::add);
+    }
+    return nodes;
+  }
+
+  @Override
+  public JsonNode fetchOneRunMetrics(String runId) throws IOException {
+
+    // Unclear from ULTIMA if this endpoint is permanently EOL
+    // Do not use in meantime as not supported
```
avarsava: Confused by this comment. Looks like you say 'do not use' then immediately use it?

grapigeau: Oh that is confusing, I never use fetchOneRunMetrics

avarsava: Maybe worth removal then

grapigeau: Yeah, I was really just holding out hope Ultima would get back to me about why this endpoint disappeared

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClientImpl.java
grapigeau: Added

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: Removed from metrics

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: Modified to throw error

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java
grapigeau: So right now the processor probably would be fine with a gross path b/c it doesn't care about anything except the run ID. It will need to once Google Bucket support is implemented.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: Renamed both the list and the function to be more descriptive

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
```diff
@@ -0,0 +1,80 @@
...
+
+import ca.on.oicr.gsi.runscanner.dto.type.Platform;
+import java.util.Objects;
+
+public class UltimaNotificationDto extends NotificationDto {
+
+  private int completedFlowNum;
+  private double readLength;
+  private int runFlows;
+  private int waferShelf;
```
djcooke, line 12: Isn't this the sequencer position? There's already a field for that.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
```diff
@@ -0,0 +1,80 @@
+package ca.on.oicr.gsi.runscanner.dto;
+
+import ca.on.oicr.gsi.runscanner.dto.type.Platform;
+import java.util.Objects;
+
+public class UltimaNotificationDto extends NotificationDto {
+
+  private int completedFlowNum;
+  private double readLength;
+  private int runFlows;
```
djcooke: Assuming these are what I'm guessing, I think renaming would add clarity and consistency. "runFlows" is especially unclear.

- `runFlows` -> `expectedFlows`
- `completedFlowNum` -> `completedFlows`

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: I suppose it is a pretty safe assumption that the timezone from the API and the one the user is configuring should be the same

avarsava: I don't think it is, but the one the user configured is the one they want to see
So... I guess actually you need to apply the API timezone and then convert to the config tz?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -0,0 +1,10 @@
+package ca.on.oicr.gsi.runscanner.scanner.processor;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import java.io.IOException;
+import java.util.List;
+
+public interface UltimaApiClient {
```
djcooke: Is this a useful interface? Why would there be other implementations?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClientImpl.java
```diff
@@ -0,0 +1,69 @@
...
+        .setReadTimeout(Duration.ofSeconds(5))
+        .build();
+  }
+
+  @Override
+  public List<JsonNode> fetchAllRunSummaries() throws IOException {
+
+    String datetimeRestrictions =
+        "from=01-01-2025&to="
+            + DateTimeFormatter.ofPattern("MM-dd-yyyy", Locale.ENGLISH).format(LocalDateTime.now());
```
djcooke: Is the `to` parameter required, or could it be excluded and the URL be made static?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: The start and completion times are _Instant_ so I don't think there should any need to convert to the config tz

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClientImpl.java
```diff
@@ -0,0 +1,69 @@
...
+    List<JsonNode> nodes = new ArrayList<>();
+    if (response.getBody() != null && response.getBody().isArray()) {
+      response.getBody().forEach(nodes::add);
+    }
+    return nodes;
+  }
+
+  /** Helper to read the token from disk and build the Auth header. */
+  private HttpHeaders getHeaders() throws IOException {
+    String token = Files.readString(Paths.get(tokenPath)).trim();
```
djcooke: Read the token once and store it in a field rather than reading the file for every request. Could probably store the HttpHeaders entirely in a field and reuse them too.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+   */
+  private static String fetchNexusApiTokenFile(ObjectNode parameters) {
+    // TODO Should read and cache token here?
+    if (parameters.hasNonNull("nexusApiTokenFile")) {
+      return parameters.get("nexusApiTokenFile").asText();
+    } else {
+      log.error("No Nexus API Token configured for Ultima");
+      return null;
+    }
+  }
```
djcooke, line 76: It looks like if either of these are missing, it will currently create the processor and show the config as valid in the UI, but then throw NPEs whenever it tries to scan. Is that right? Can it be shown as invalid instead?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultUltima extends RunProcessor {
+  private static final LatencyHistogram api_fetch_time =
+      new LatencyHistogram(
+          "runscanner_ultima_scan_completeness_time", "The time to query the NEXUS API.");
```
djcooke: I don't think this is "scan completeness time." Something like `runscanner_ultima_fetch_runs_time` would be more accurate and clear. Specifying that it's for fetching all runs is ideal because if we need to make another API call in the future, it's unlikely we'd want to combine them into the same metric.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  /**
+   * Instead of scanning disk, we fetch the list of Run IDs from the API. We return "Virtual" File
+   * objects representing each Run ID.
+   */
+  @Override
+  public Stream<File> getRunsFromRoot(File root) {
+
+    try (AutoCloseable latency = api_fetch_time.start()) {
```
djcooke: You're including more than the API fetch time in this block. The metric might be better scoped inside the `UltimaApiClient` class.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+      return allRunInfo
+          .stream()
+          .flatMap(
+              node -> {
+                String runId = node.path("runid").asText();
+
+                // Skip if runId is null, empty, or just whitespace
+                if (runId == null || runId.trim().isEmpty()) {
+                  return Stream.empty();
+                }
```
djcooke, line 96: Is this something you observed? Seems like a problem with the API that we should bring up if so?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+
+                runCache.put(runId, node); // Cache for the process() step
+
+                // TODO eventually will be google bucket path with runID in subfolder name
+                return Stream.of(new File(root, runId));
```
djcooke: I assume this comment means that for OICR it will be a Google bucket? Will other options be supported?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+  @Override
+  public Stream<File> getRunsFromRoot(File root) {
+
+    try (AutoCloseable latency = api_fetch_time.start()) {
+
+      List<JsonNode> allRunInfo = apiClient.fetchAllRunSummaries();
+
+      return allRunInfo
+          .stream()
+          .flatMap(
```
djcooke: You're flatmapping to streams of size 0-1. Can map to File, and filter nulls instead if nulls are an actual possibility, but I'm hoping they're not.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+    double numReadPass = json.path("num_reads_pass_filter").asDouble();
+    double numBeads = json.path("numbeads").asDouble();
+
+    ObjectNode runMetrics = metrics.addObject();
+    runMetrics.put("numBeads", numBeads);
+    runMetrics.put("outputReads", Long.parseLong(json.path("pf_output_reads").asText("0")));
+    if (numBeads > 0) {
+      runMetrics.put("passFilterPercent", numReadPass / numBeads);
+    }
+    dto.setMetrics(mapper.writeValueAsString(metrics));
```
djcooke: This is not consistent with metric formats from Run Scanner that are displayed in MISO. There is a "table" metric type that you likely want to use.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+
+    dto.setSequencerPosition(json.path("Chuck").asText());
```
djcooke, line 156: Ok, the "chuck" is the position on the sequencer that it runs in? Then the wafer shelf is where it's stored when queued? Is that useful for something?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+    dto.setContainerSerialNumber(wafer);
+    dto.setContainerModel(extractModel(wafer));
+
+    int status = json.path("runstatus").asInt();
+    int completed = json.path("isruncompleted").asInt();
+    String errmsg = json.path("errormsg").asText();
+    HealthType health = translateStatus(status, completed, errmsg);
+    dto.setHealthType(health);
+
+    // TODO better way for timezone?
```
djcooke: Resolve TODO

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+    String zoneString = json.path("timezone").asText();
+    LocalDateTime ldt =
+        LocalDateTime.parse(json.path("startdatetime").asText(), NEXUS_DATE_FORMATTER);
+    Instant startDate = extractTime(ldt, zoneString);
+    dto.setStartDate(startDate);
+    if (health == HealthType.COMPLETED) {
+      Duration runtime = Duration.ofMinutes(json.path("runtime").asInt());
+      dto.setCompletionDate(calculateCompletedDate(startDate, runtime));
+    }
```
djcooke: Failed and stopped runs should also have a completion date. Check `health.isDone()` instead of the specific status

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+
+    } catch (Exception e) {
+      log.error("Invalid timezone cannot determine start or completion time.");
+      return null;
+    }
+  }
+
+  public String extractModel(String serialNumber) throws IOException {
+    // Example: FC12.2T0123456789 -> FTC12.2T
+    Pattern pattern = Pattern.compile("^(.{4}\\..{2})");
```
djcooke, line 227: A little concerned for forwards compatibility, but I'm guessing there's not a better option. Maybe we could ask for the model to be provided in the API and then update Run Scanner to use that later if it happens?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,228 @@
...
+    if (matcher.find()) {
+      // Group 1 contains the extracted text
+      return matcher.group(1);
+    } else {
+      throw new IOException("Cannot determine containerModel");
+    }
+  }
+
+  public HealthType translateStatus(int status, int completed, String errmsg) {
+    if (status == 100 && completed == 1 && errmsg.isEmpty()) {
```
djcooke: What does status 100 mean? What other status values are there?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClientImpl.java
```diff
@@ -0,0 +1,69 @@
...
+
+    HttpEntity<String> entity = new HttpEntity<>(getHeaders());
+
+    ResponseEntity<JsonNode> response =
+        restTemplate.exchange(url, HttpMethod.GET, entity, JsonNode.class);
+
+    List<JsonNode> nodes = new ArrayList<>();
+    if (response.getBody() != null && response.getBody().isArray()) {
+      response.getBody().forEach(nodes::add);
+    }
```
djcooke: Looks like this is ignoring and returning nothing if the response doesn't look the way we're expecting?

scanner/src/test/resources/ultima/123456/reference.json
```diff
@@ -0,0 +1,18 @@
+{
+  "runAlias": "123456",
+  "sequencerName": "v987",
+  "software": null,
```
djcooke, line 4: Shouldn't this include the recipe version? ...though I'd actually argue that's probably not a "software version" anyway?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
grapigeau: sequencerPosition is "Chuck" (Can be 1 or 2 and describes the position where the wafer is being imaged), waferShelf (1-6 and describes from which position the wafer was retrieved) is different

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClientImpl.java
grapigeau: the to and from is required for the `/runsummary` endpoint

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: Status is 0-100, it's percentage completion

scanner/src/test/resources/ultima/123456/reference.json
grapigeau: I could include the SequencingRecipe which is something like _UG_116cycles_Baseline_1.8.0.0_ if that's a better fit for "software"

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: I can add that to my requests for Ultima, would be nice to have it in its own field I agree

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
grapigeau: It is useful for testing

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: No I haven't seen this, was really a "just in case" check.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: I think the idea is to also support local (or locally mounted) file systems

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: I don't know what but Savo added it to his list of values to include

scanner/src/test/resources/ultima/123456/reference.json
grapigeau: OH nevermind, software can be set but for testing it's made "null" regardless in _AbstractProcessorTest_

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
djcooke: Could you add a note of that please? And is "completed" is supposed to mean "completed successfully" or just "no longer running?" That will affect the logic here.

Below, status == 0 will be reported as `UNKNOWN`. Are you sure it shouldn't be `RUNNING`?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClientImpl.java
djcooke: Gross. Hopefully time zones match. Is it possible to set a date in the future just in-case? +24hrs maybe? Making these optional would be a good request for Ultima too.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
djcooke: Mocking via Mockito would be better for testing, but this is ok.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Configuration.java
```diff
@@ -41,7 +47,11 @@ public boolean isValid() {
 
     // If the path is VIRTUAL, we only care that a path string and timezone are provided
     if (processor != null && processor.getPathType() == PathType.VIRTUAL) {
-      return path != null && timeZone != null;
+      return path != null
+          && timeZone != null
+          && parameters.hasNonNull("nexusApiAddress")
+          && parameters.hasNonNull("nexusApiTokenFile")
+          && new File(parameters.get("nexusApiTokenFile").asText()).canRead();
```
djcooke: It makes sense to use path type to determine whether to check permissions, but if we add another processor with `PathType.VIRTUAL`, the parameter checks here won't make sense. Could add a `validateParameters` method to `RunProcessor` with a default implementation that does nothing, and call that here. Then override that in `DefaultUltima` to check the parameters it requires.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -127,25 +121,37 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+    values
+        .addObject()
+        .put("name", "outputReads")
+        .put("value", json.path("pf_output_reads").asText("0"));
     if (numBeads > 0) {
-      runMetrics.put("passFilterPercent", numReadPass / numBeads);
+      values
+          .addObject()
+          .put("name", "passFilterPercent")
+          .put("value", String.format("%.2f %%", (numReadPass / numBeads) * 100));
```
djcooke: Could give these nicer names since they're for display purpose

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: isruncompleted is completed successfully. I can change the var from "completed" to "succeeded"

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClientImpl.java
grapigeau: Modified so it throws an error - that way it won't be failing silently

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -139,16 +139,16 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
 
     values
         .addObject()
-        .put("name", "numBeads")
+        .put("name", "Num Beads")
```
djcooke: "Number of"?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -211,12 +211,12 @@ public String extractModel(String serialNumber) throws IOException {
...
+    if (status == 100 && succeeded == 1 && errmsg.isEmpty()) {
       return HealthType.COMPLETED;
-    } else if ((status == 100 && completed == 0) || (completed == 0 && !errmsg.isEmpty())) {
+    } else if ((status == 100 && succeeded == 0) || (succeeded == 0 && !errmsg.isEmpty())) {
       return HealthType.FAILED;
-    } else if (status > 0 && errmsg.isEmpty()) {
+    } else if (status >= 0 && errmsg.isEmpty()) {
       return HealthType.RUNNING;
     } else {
       return HealthType.UNKNOWN;
```
djcooke: Is `status` always 100 if the run has stopped? e.g. if the run starts and fails immediately?

There are conditions that you're missing, which may or may not be possible, but it's not obvious from looking at this. Simplifying might be better either way:

```
if !errmsg.empty
  FAILED
else if status == 100
  succeeded ? COMPLETED : FAILED
else if status >= 0
  RUNNING
else
  UNKNOWN
```

grapigeau: If the run has stopped the status is whatever percentage it was at when the stop occurred.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
```diff
@@ -0,0 +1,80 @@
...
+
+  @Override
+  public String toString() {
+    return super.toString()
+        + ", UltimaNotificationDto ["
+        + "completedFlows="
+        + completedFlows
+        + "expectedFlows="
+        + expectedFlows
+        + "readLength="
```
callunity: Do these need `, ` at the start of the strings for better comprehension?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,241 @@
...
+
+    values
+        .addObject()
+        .put("name", "Number of Beads")
+        .put("value", new BigDecimal(numBeads).toPlainString()); // prevent scientific notation
+    values
+        .addObject()
+        .put("name", "Output Reads")
+        .put("value", json.path("pf_output_reads").asText("0"));
+    if (numBeads > 0) {
```
callunity: What do you think about setting Pass Filter % to 0 if numBeads is 0 ? Means anything consuming the metrics object doesn't have to account for the fact that one metric may be entirely absent in some cases (undocumented).

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,241 @@
...
+    metrics.add(chartNode);
+    dto.setMetrics(mapper.writeValueAsString(metrics));
+
+    dto.setSequencerPosition(json.path("Chuck").asText());
+
+    String wafer = json.path("wafer").asText();
+    dto.setContainerSerialNumber(wafer);
+    dto.setContainerModel(extractModel(wafer));
+
+    int status = json.path("runstatus").asInt(); // Percentage 0-100
```
callunity: `status` -> `pctCompleted` ? I think that also clarifies the checks in translateStatus as well

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,241 @@
...
+    // No paired ends
+    dto.setPairedEndRun(false);
+
+    return dto;
+  }
+
+  public Instant extractTime(LocalDateTime datetime, String timeZone) {
+    try {
+      ZoneId zone = ZoneId.of("UTC");
+      if (timeZone.equals("Eastern Standard Time")) {
```
callunity: Where does EST come from? The instrument itself?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,241 @@
...
+
+  public String extractModel(String serialNumber) throws IOException {
+    // Example: FC12.2T0123456789 -> FTC12.2T
+    Pattern pattern = Pattern.compile("^(.{4}\\..{2})");
+    Matcher matcher = pattern.matcher(serialNumber);
+    if (matcher.find()) {
+      // Group 1 contains the extracted text
+      return matcher.group(1);
+    } else {
+      throw new IOException("Cannot determine containerModel");
```
callunity: Would debugging be easier if the serial number was included in the exception message?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -0,0 +1,86 @@
...
+        "from=01-01-2025&to="
+            + DateTimeFormatter.ofPattern("MM-dd-yyyy", Locale.ENGLISH)
+                .format(LocalDateTime.now().plusDays(1));
+    String url = String.format("%s/api/data/runsummary?%s", apiUrl, datetimeRestrictions);
+
+    HttpEntity<String> entity = new HttpEntity<>(headers);
+
+    ResponseEntity<JsonNode> response;
+    try (AutoCloseable latency = all_run_api_fetch_time.start()) {
+      response = restTemplate.exchange(url, HttpMethod.GET, entity, JsonNode.class);
```
callunity, line 80: Worth checking for a 200-series response code?

RELEASE_NOTES.md
```diff
@@ -3,6 +3,7 @@
 Changes:
 
  * Update read the doc links
+ * Added support for Ultima UG 100
```
callunity: What do you think about qualifying this? "Added support for Ultima UG 100 run information" since we haven't yet implemented the analysis information part

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,241 @@
...
+        .put("value", new BigDecimal(numBeads).toPlainString()); // prevent scientific notation
+    values
+        .addObject()
+        .put("name", "Output Reads")
+        .put("value", json.path("pf_output_reads").asText("0"));
+    if (numBeads > 0) {
+      values
+          .addObject()
+          .put("name", "Pass Filter %")
+          .put("value", String.format("%.2f %%", (numReadPass / numBeads) * 100));
```
callunity: What do you think about just listing the percentage number and omitting the terminal ` %` ? For ease of computational processing downstream

grapigeau: I'm not opposed - this was just based on Illumina formatting for the metrics objects

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: Yes from the instrument

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
```diff
@@ -18,7 +21,11 @@ public boolean equals(Object obj) {
     UltimaNotificationDto other = (UltimaNotificationDto) obj;
 
     return Objects.equals(completedFlows, other.completedFlows)
-        && Objects.equals(waferShelf, other.waferShelf);
+        && Objects.equals(expectedFlows, other.expectedFlows)
+        && Objects.equals(waferShelf, other.waferShelf)
+        && Objects.equals(sequencingStatus, other.sequencingStatus)
+        && Objects.equals(analysisStatus, other.analysisStatus)
+        && Objects.equals(uploadStatus, other.uploadStatus);
   }
```
callunity: Should these be here? We expect that status fields will be mutable, and differences shouldn't indicate a need to represent them as different objects.

grapigeau: The only fields that aren't going to change over the course of run detection are expectedFlows and Wafershelf so I will just include those.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
callunity: I can understand how you got there! But Illumina has some esoteric choices which we don't necessarily need to replicate.

docs/appendices.md
```diff
@@ -95,192 +112,245 @@ esp bc they might vary in meaning b/w PromION and MinION -->
...
+    </tr>
+    <tr>
+        <td>completedFlows</td>
+        <td></td>
+        <td></td>
+        <td></td>
+        <td>✓</td>
+    </tr>
+    <tr>
+        <td>wafershelf</td>
```
avarsava: Update?

docs/appendices.md
```diff
@@ -324,6 +394,13 @@ unsupported.
 * iSeq
 * MiniSeq
 
+#### Ultima - "default" processor
+* *UG 100*
+
+**Notes for Ultima Sequencers**
+* The sequencing, analysis, and upload status are recorded as ints where 0 is not started, 
+  1 is in progress, 2 is complete, and 3+ is error
```
avarsava: Should this be an enum instead? We don't need to export everything Ultima does exactly.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: The API runsummary gives values as:
`"timezone": "Eastern Standard Time",
 "startdatetime": "2024-05-07 10:00:00",`

callunity: Since our time zone switches between "Eastern Standard Time" and "Eastern Daylight Time", might be worth seeing if we can automate this - looks like `TimeZone.getTimeZone(timeZone).toZoneId()` falls back to UTC if it doesn't find a match for whatever timezone string the sequencer is serving, which I think should do what we need?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -160,10 +159,20 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
     dto.setContainerSerialNumber(wafer);
     dto.setContainerModel(extractModel(wafer));
 
-    int status = json.path("runstatus").asInt(); // Percentage 0-100
-    int succeeded = json.path("isruncompleted").asInt(); // Successful run
+    int pctCompleted = json.path("runstatus").asInt(); // Percentage 0-100
+    int succeeded = json.path("isruncompleted").asInt(); // Successful run 0 or 1
     String errmsg = json.path("errormsg").asText();
-    HealthType health = translateStatus(status, succeeded, errmsg);
+    String analAndUpStatus = json.path("analysisstatus").asText();
```
callunity, line 169: it looks like `analysisstatus` is a 4-digit number - do you know what the first and last characters represent?

grapigeau: Ultima simply told us to ignore them

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,264 @@
...
+    return startdate.plus(runtime);
+  }
+
+  @Override
+  public PathType getPathType() {
+    return PathType.VIRTUAL;
+  }
+
+  @Override
+  public boolean validateParameters(ObjectNode parameters) {
```
avarsava, line 274: Not sure how Jackson handles empty string - do you also need to check parameters.get("whatever").isBlank()?

docs/appendices.md
```diff
@@ -81,6 +87,17 @@
                 <td>(no additional parameters)</td>
                 <td></td>
             </tr>
+            <tr colspan="2">
+                <th>ULTIMA</th>
+            </tr>
+            <tr>
+                <td>nexusApiAddress (required)</td>
+                <td>The base URL of the Nexus API</td>
```
avarsava: Might want to provide an example, given that your methods have the `/api` part baked in, might not be clear to a user whether the 'API Address' should include that or not

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Configuration.java
```diff
@@ -58,26 +69,34 @@ public String validitySummary() {
     }
     if (processor == null) summary += "Processor is null! ";
     if (timeZone == null) summary += "TimeZone is null!";
```
avarsava:
```suggestion
    if (timeZone == null) summary += "TimeZone is null! ";
```

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Configuration.java
```diff
@@ -58,26 +69,34 @@ public String validitySummary() {
...
+    if (processor != null && processor.getPathType() == PathType.VIRTUAL) {
+      if (!processor.validateParameters(parameters))
+        summary += "Parameters not configured correctly!";
```
avarsava:
```suggestion
        summary += "Parameters not configured correctly! ";
```

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
callunity: I see.... could you add a comment here about how the first and last characters are ignored, as per Ultima's guidance, and the characters at [1] and [2] represent analysis status and upload status?

docs/appendices.md
callunity: Seconding this!

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -228,18 +235,39 @@ private String extractModel(String serialNumber) throws IOException {
...
+    switch (status) {
+      case 0:
+        return UltimaProcessStatus.PENDING;
+      case 1:
+        return UltimaProcessStatus.RUNNING;
+      case 2:
+        return UltimaProcessStatus.COMPLETE;
+    }
+    return UltimaProcessStatus.FAILED;
+  }
```
callunity: What do you think about moving this method to the UltimaProcessStatus enum? (Unless the translation implementation could be different in a different class?)

docs/appendices.md
callunity: -> `waferShelf` ?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -0,0 +1,89 @@
...
+      throw new IOException(e);
+    }
+    List<JsonNode> nodes = new ArrayList<>();
+    if (!response.getStatusCode().is2xxSuccessful()) {
+      throw new IOException("API request failed with status code: " + response.getStatusCode());
+    }
+    if (response.getBody() != null && response.getBody().isArray()) {
+      response.getBody().forEach(nodes::add);
+    } else {
+      throw new IOException("Unexpected JSON response format from API request.");
```
callunity: What do you think about including the response body in the IOExceptions for easier debugging?
And, what do you think about incrementing a runscanner_nexus_failed_requests counter each time it throws? This will make for easier alerting on communication issues with the Nexus server.

scanner/src/test/resources/ultima/123456/reference.json
```diff
@@ -0,0 +1,22 @@
...
+  "waferShelf": 1,
+  "readLength":300.0,
+  "metrics": "[{\"type\":\"chart\",\"values\":[{\"name\":\"Number of Beads\",\"value\":\"600000000\"},{\"name\":\"Output Reads\",\"value\":\"450000000\"},{\"name\":\"Pass Filter %\",\"value\":\"83.33\"}]}]",
+  "sequencerPosition": "1",
+  "containerSerialNumber": "FC12.2T1234567890",
+  "containerModel": "FC12.2T",
+  "healthType": "COMPLETED",
+  "analysisStatus": 2,
+  "uploadStatus": 2,
+  "sequencingStatus": 2,
```
callunity: Shouldn't these be represented as the enums now? (It's concerning that the tests are still passing even though this didn't get changed...)

grapigeau: The test uses the equals function within the class.
So maybe it is best to have it set so that the equals and hash account for the statuses?

callunity: What if we added an `if UltimaNotificationDto.class` clause to the [AbstractProcessorTest](https://github.com/miso-lims/runscanner/blob/2e3c8e7a4f079e4cfcec4762908d4baf04af50b4/scanner/src/test/java/ca/on/oicr/gsi/runscanner/scanner/processor/AbstractProcessorTest.java#L59) which compares the status fields?

docs/appendices.md
```diff
@@ -324,6 +415,13 @@ unsupported.
 * iSeq
 * MiniSeq
 
+#### Ultima - "default" processor
+* *UG 100*
+
+**Notes for Ultima Sequencers**
+
+* Current support is in development and subject to change
```
avarsava, line 423: See lines 383-385. The 'UG 100' item should be italicized instead of bold if this is the case.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -0,0 +1,280 @@
...
+
+  @Override
+  public PathType getPathType() {
+    return PathType.VIRTUAL;
+  }
+
+  @Override
+  public boolean validateParameters(ObjectNode parameters) {
+    return parameters.hasNonNull("nexusApiAddress")
+        && !parameters.get("nexusApiAddress").asText().isEmpty()
```
avarsava: String.isBlank() will cover more error cases than isEmpty() 
https://www.baeldung.com/java-string-isempty-vs-isblank#comparing-isblank-vs-isempty

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -0,0 +1,99 @@
...
+  public UltimaApiClient(String apiUrl, String tokenPath) {
+    this.apiUrl = apiUrl;
+    this.restTemplate =
+        new RestTemplateBuilder()
+            .setConnectTimeout(Duration.ofSeconds(5))
+            .setReadTimeout(Duration.ofSeconds(5))
+            .build();
+
+    try {
+      String token = Files.readString(Paths.get(tokenPath)).trim();
```
avarsava, line 49: Do you want to check early whether the token is blank?

docs/appendices.md
avarsava: Oh, actually, apparently single asterisks and underscore are the same in markdown! Nevermind, but this note isn't needed if the item is italicized

grapigeau: Good to know about markdown! Mike did reccomend adding in an explicit note

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
grapigeau: At this point it's already been checked that it's a readable file so if the token blank and the auth fails that error status should be returned ( same as if the token is invalid or typed wrong ) which seems sufficient to me

avarsava (approved):

callunity (approved):

PR #204, opened 2026-05-15
fix ultima parameter config validation

grapigeau:
JIRA Ticket:

- [x] Updates release notes

### DIFF
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
@@ -35,11 +35,16 @@ public class DefaultUltima extends RunProcessor {
   private final UltimaApiClient apiClient;
 
   public static DefaultUltima create(Builder builder, ObjectNode parameters) {
-    return new DefaultUltima(
-        builder, fetchNexusApiUrl(parameters), fetchNexusApiTokenFile(parameters));
+    try {
+      return new DefaultUltima(
+          builder, fetchNexusApiUrl(parameters), fetchNexusApiTokenFile(parameters));
+    } catch (IOException e) {
+      log.error("Could not create Ultima run processor: {}", e.getMessage(), e);
+      return null;
+    }
   }
 
-  protected DefaultUltima(Builder builder, String apiUrl, String tokenPath) {
+  protected DefaultUltima(Builder builder, String apiUrl, String tokenPath) throws IOException {
     super(builder);
     this.apiClient = new UltimaApiClient(apiUrl, tokenPath);
   }
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
@@ -4,7 +4,6 @@ import ca.on.oicr.gsi.runscanner.scanner.LatencyHistogram;
 import com.fasterxml.jackson.databind.JsonNode;
 import io.prometheus.metrics.core.metrics.Counter;
 import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.time.Duration;
@@ -14,7 +13,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import org.springframework.boot.web.client.RestTemplateBuilder;
-import org.springframework.context.annotation.Bean;
 import org.springframework.http.HttpEntity;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpMethod;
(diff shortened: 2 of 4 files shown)


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -53,7 +56,8 @@ public UltimaApiClient(String apiUrl, String tokenPath) {
       this.headers = headers;
     } catch (IOException e) {
       // config should have been invalid by this point
-      throw new UncheckedIOException("Failed to read Ultima API token from " + tokenPath, e);
+      log.error("Could not initialize Ultima API client: " + e);
```
djcooke: Whenever including an exception in the log, use the version of `log.error` that takes the Exception as a separate argument so that a proper stack trace is logged.

In this case, I think `UltimaApiClient` should throw the exception though. The caller (processor) should decide what to do if that happens since it's not capable of functioning and should be aware of that.

callunity (approved):

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -272,9 +274,23 @@ public PathType getPathType() {
...
+            && new File(parameters.get("nexusApiTokenFile").asText()).canRead();
+
+    if (!basicChecks) {
+      return false;
+    }
+
+    try {
+      String path = parameters.get("nexusApiTokenFile").asText();
+      Files.readString(Paths.get(path)).trim();
+      return true;
```
djcooke: There shouldn't be a need to read this file more than once. I don't think it should need to happen during validation since you're making construction handle the error already. Checking that there's no `IOException` at this point doesn't mean that there won't be one later anyway.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
grapigeau: My thinking was the error shouldn't interfere with the processor creation since having one Builder header fail shouldn't matter to the whole process? 
But you're thinking [RunProcessor](https://github.com/miso-lims/runscanner/blob/master/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/RunProcessor.java) should handle it?

djcooke: Yes. `UltimaApiClient` cannot function at all if this fails. If that non-functionality is handled inside `UltimaApiClient`, then it is unclear to its consumer (the processor) That means that the processor should treat it as functional, which means it will get errors or unexpected behaviour every time it tries to do anything with the API, and the log will be spammed with the same errors/exceptions over and over. Or worse - it will think it's working and data will be bad or incomplete.

If an exception is thrown here, then the consumer (processor) can decide how to handle that, such as reporting itself invalid/erroring in the UI, and not continuing to attempt to use the API which it knows will not work.

djcooke: The other option is to read the file in the processor and pass that to the API class constructor instead so that this doesn't need to be considered here at all.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -34,27 +33,22 @@ public class UltimaApiClient {
           .register();
 
   private final String apiUrl;
-  private final HttpHeaders headers;
   private final RestTemplate restTemplate;
+  private HttpHeaders headers;
```
djcooke: Should still be `final`

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -67,6 +61,12 @@ public RestTemplate restTemplate() {
 
   public List<JsonNode> fetchAllRunSummaries() throws IOException {
 
+    if (this.headers == null) {
```
djcooke: How could `headers` be null?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
djcooke: Didn't notice this before, but this isn't doing anything since it's not in a `@Configuration` class or similar that are wired when Spring starts up. You're also creating the `RestTemplate` in the constructor, so this isn't needed anyway.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
grapigeau: I really don't think it should be, because it shouldn't be reaching that point if the headers threw an error but I wanted to be doubly sure

djcooke: It's not possible - it's `final` now and only ever set to one value that can't be null. Handling nulls where they shouldn't be is often safer, but for a private final field it's unnecessary and a bit confusing.

djcooke (approved):

callunity (approved):

PR #207, opened 2026-06-02
add Sample DB support for Ultima runs

grapigeau:
add consumables to ultima runs

JIRA Ticket: https://jira.oicr.on.ca/browse/GLT-4804 and https://jira.oicr.on.ca/browse/GLT-4807

- [x] Updates release notes

### DIFF
--- a/docs/appendices.md
+++ b/docs/appendices.md
@@ -92,11 +92,19 @@
             </tr>
             <tr>
                 <td>nexusApiAddress (required)</td>
-                <td>The base URL of the Nexus API (Ex. https://example-nexus.com)</td>
+                <td>The base URL of the Nexus API (Ex. https://example-nexus.ca)</td>
             </tr>
             <tr>
                 <td>nexusApiTokenFile (required)</td>
                 <td>The location of a file containing a Nexus Auth Token</td>
+            </tr>
+            <tr>
+                <td>sampleDBApiAddress (optional)</td>
+                <td>The base URL of the Sample DB API (Ex. https://sample-db.ca)</td>
+            </tr>
+            <tr>
+                <td>sampleDBApiTokenFile (optional)</td>
+                <td>The location of a file containing a Sample DB Auth Token</td>
             </tr>
         </table></td>
     </tr>
@@ -213,13 +221,6 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
     </tr>
     <tr>
-        <td>poolNames</td>
-        <td>✓</td>
-        <td></td>
-        <td></td>
-        <td></td>
-    </tr>
-    <tr>
         <td>runBasesMask</td>
         <td></td>
         <td>✓</td>
@@ -246,6 +247,13 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
         <td></td>
         <td></td>
+    </tr>
+    <tr>
+        <td>consumables</td>
+        <td></td>
+        <td>✓</td>
+        <td></td>
+        <td>✓</td>
     </tr>
     <tr>
         <td>imgCycle</td>
@@ -277,10 +285,10 @@ esp bc they might vary in meaning b/w PromION and MinION -->
     </tr>
     <tr>
         <td>poolNames</td>
-        <td></td>
+        <td>✓</td>
         <td>✓</td>
         <td></td>
-        <td></td>
+        <td>✓</td>
     </tr>
     <tr>
         <td>readLength</td>
--- a/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
+++ b/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
@@ -2,7 +2,9 @@ package ca.on.oicr.gsi.runscanner.dto;
 
 import ca.on.oicr.gsi.runscanner.dto.type.Platform;
 import ca.on.oicr.gsi.runscanner.dto.type.UltimaProcessStatus;
+import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 public class UltimaNotificationDto extends NotificationDto {
 
@@ -13,6 +15,8 @@ public class UltimaNotificationDto extends NotificationDto {
   private UltimaProcessStatus sequencingStatus;
   private UltimaProcessStatus analysisStatus;
   private UltimaProcessStatus uploadStatus;
+  private List<Consumable> consumables;
+  private List<String> poolNames;
 
   @Override
   public boolean equals(Object obj) {
@@ -82,6 +86,22 @@ public class UltimaNotificationDto extends NotificationDto {
     return uploadStatus;
   }
 
+  public void setConsumables(List<Consumable> consumables) {
+    this.consumables = consumables;
+  }
+
+  public List<Consumable> getConsumables() {
+    return consumables;
+  }
+
+  public void setPoolNames(List<String> poolNames) {
+    this.poolNames = poolNames;
+  }
+
+  public List<String> getPoolNames() {
+    return poolNames;
+  }
+
   public void setUploadStatus(UltimaProcessStatus uploadStatus) {
     this.uploadStatus = uploadStatus;
   }
@@ -92,13 +112,25 @@ public class UltimaNotificationDto extends NotificationDto {
   }
 
   @Override
+  public Optional<String> getLaneContents(int lane) {
+    if (poolNames.size() == 1) {
+      return Optional.of(poolNames.get(0));
+    }
+    return Optional.empty();
+  }
+
+  @Override
   public String toString() {
     return super.toString()
         + ", UltimaNotificationDto ["
         + ", expectedFlows="
         + expectedFlows
+        + ", consumables="
+        + consumables
         + ", wafershelf="
         + waferShelf
-        + "]";
+        + ", poolNames=["
+        + poolNames
+        + "]]";
   }
 }
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
@@ -1,5 +1,6 @@
 package ca.on.oicr.gsi.runscanner.scanner.processor;
 
+import ca.on.oicr.gsi.runscanner.dto.Consumable;
 import ca.on.oicr.gsi.runscanner.dto.NotificationDto;
 import ca.on.oicr.gsi.runscanner.dto.UltimaNotificationDto;
 import ca.on.oicr.gsi.runscanner.dto.type.HealthType;
@@ -37,16 +38,27 @@ public class DefaultUltima extends RunProcessor {
   public static DefaultUltima create(Builder builder, ObjectNode parameters) {
     try {
       return new DefaultUltima(
-          builder, fetchNexusApiUrl(parameters), fetchNexusApiTokenFile(parameters));
+          builder,
+          fetchNexusApiUrl(parameters),
+          fetchNexusApiTokenFile(parameters),
+          fetchOptionalParameter(parameters, "sampleDBApiAddress"),
+          fetchOptionalParameter(parameters, "sampleDBApiTokenFile"));
     } catch (IOException e) {
       log.error("Could not create Ultima run processor: {}", e.getMessage(), e);
       return null;
     }
   }
 
-  protected DefaultUltima(Builder builder, String apiUrl, String tokenPath) throws IOException {
+  protected DefaultUltima(
+      Builder builder,
+      String apiUrlNexus,
+      String tokenPathNexus,
+      String apiUrlSampleDB,
+      String tokenPathSampleDB)
+      throws IOException {
     super(builder);
-    this.apiClient = new UltimaApiClient(apiUrl, tokenPath);
+    this.apiClient =
+        new UltimaApiClient(apiUrlNexus, tokenPathNexus, apiUrlSampleDB, tokenPathSampleDB);
   }
 
   protected DefaultUltima(Builder builder, UltimaApiClient apiClient) {
@@ -56,7 +68,7 @@ public class DefaultUltima extends RunProcessor {
 
   /**
    * @param parameters ObjectNode
-   * @return String with base URL for the ULTIMA API
+   * @return String with base URL for the ULTIMA Nexus API
    */
   private static String fetchNexusApiUrl(ObjectNode parameters) {
     if (parameters.hasNonNull("nexusApiAddress")) {
@@ -69,13 +81,26 @@ public class DefaultUltima extends RunProcessor {
 
   /**
    * @param parameters ObjectNode
-   * @return String filename where token is stored
+   * @return String filename where Nexus token is stored
    */
   private static String fetchNexusApiTokenFile(ObjectNode parameters) {
     if (parameters.hasNonNull("nexusApiTokenFile")) {
       return parameters.get("nexusApiTokenFile").asText();
     } else {
       log.error("No Nexus API Token configured for Ultima, this config should be invalid");
+      return null;
+    }
+  }
+
+  /**
+   * @param parameters ObjectNode
+   * @param parameterName String
+   * @return The value stored at the parameter name or null if it doesn't exist
+   */
+  private static String fetchOptionalParameter(ObjectNode parameters, String parameterName) {
+    if (parameters.hasNonNull(parameterName)) {
+      return parameters.get(parameterName).asText();
+    } else {
       return null;
     }
   }
@@ -124,6 +149,7 @@ public class DefaultUltima extends RunProcessor {
     }
 
     UltimaNotificationDto dto = new UltimaNotificationDto();
+
     dto.setRunAlias(runId);
     dto.setSequencerName(json.path("sysid").asText());
     dto.setSoftware(json.path("SequencingRecipe").asText());
@@ -208,6 +234,39 @@ public class DefaultUltima extends RunProcessor {
     // No paired ends
     dto.setPairedEndRun(false);
 
+    String ampSamplePlate = json.path("AMP_SamplePlate").asText("");
+    dto.setPoolNames(getPoolsFromSampleDB(ampSamplePlate));
+
+    List<Consumable> consumables = new ArrayList<>();
+    consumables.add(new Consumable("Amplification Sample Plate Serial Number", ampSamplePlate));
+    consumables.add(
+        new Consumable(
+            "Amplification Chilled Rack Lot Number", json.path("AMP_ChilledRack").asText()));
+    consumables.add(
+        new Consumable("Amplification RT Rack Lot Number", json.path("AMP_RTRack").asText()));
+    consumables.add(
+        new Consumable("Amplification Tube Array Lot Number", json.path("AMP_TubeArray").asText()));
+    consumables.add(
+        new Consumable(
+            "Amplification Break Container Lot Number", json.path("AMP_BreakContainer").asText()));
+    consumables.add(
+        new Consumable("Amplification Wash 1 Lot Number", json.path("AMP_Wash1").asText()));
+    consumables.add(
+        new Consumable("Amplification Wash 2 Lot Number", json.path("AMP_Wash2").asText()));
+    consumables.add(
+        new Consumable(
+            "Amplification Enrichment Bead Lot Number", json.path("AMP_EnrichmentBead").asText()));
+    consumables.add(new Consumable("Sequencing Rack Lot Number", json.path("SampleRack").asText()));
+    consumables.add(new Consumable("Sample Tube Lot Number", json.path("SampleTube").asText()));
+    consumables.add(
+        new Consumable(
+            "Sequencing Cartridge Lot Number", json.path("SequencingCartridge").asText()));
+    consumables.add(
+        new Consumable("Wash Container Lot Number", json.path("WashContainer").asText()));
+    consumables.add(new Consumable("Wafer Serial Number", json.path("Wafer").asText()));
+
+    dto.setConsumables(consumables);
+
     return dto;
   }
 
@@ -280,6 +339,31 @@ public class DefaultUltima extends RunProcessor {
   @Override
   public PathType getPathType() {
     return PathType.VIRTUAL;
+  }
+
+  private List<String> getPoolsFromSampleDB(String ampSamplePlate) {
+    List<JsonNode> pools = new ArrayList<>();
+    List<String> poolNames = new ArrayList<>();
+    try {
+      JsonNode samplePlateNode = apiClient.fetchSampleDB(ampSamplePlate);
+      if (samplePlateNode != null
+          && samplePlateNode.has("pools")
+          && samplePlateNode.get("pools").isArray()) {
+        samplePlateNode.get("pools").forEach(pools::add);
+      } else {
+        log.error("Couldn't parse response from Sample DB, no pool names set");
+      }
+
+      pools.forEach(
+          node -> {
+            if (node.has("libraryPool")) {
+              poolNames.add(node.get("libraryPool").asText());
+            }
+          });
+    } catch (IOException e) {
+      log.error("Couldn't access Sample DB, no pool names set", e);
+    }
+    return poolNames;
   }
 
   @Override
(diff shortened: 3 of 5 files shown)


docs/appendices.md
```diff
@@ -217,7 +217,7 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td>✓</td>
         <td></td>
         <td></td>
-        <td></td>
+        <td>✓</td>
```
avarsava: Please also add 'consumables' to the table

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
```diff
@@ -99,6 +118,8 @@ public String toString() {
         + expectedFlows
         + ", wafershelf="
         + waferShelf
-        + "]";
+        + ", poolNames=["
+        + poolNames
+        + "]]";
```
avarsava, line 134: Don't want to add Consumables to the string representation?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -80,6 +92,32 @@ private static String fetchNexusApiTokenFile(ObjectNode parameters) {
     }
   }
 
+  /**
+   * @param parameters ObjectNode
+   * @return String with base URL for the ULTIMA Sample DB API
+   */
+  private static String fetchSampleDBApiUrl(ObjectNode parameters) {
+    if (parameters.hasNonNull("sampleDBApiAddress")) {
```
avarsava: New parameters need to go into the parameters table in the doc too

docs/appendices.md
avarsava: also... for some reason, 'poolNames' is listed twice on this table, once for PacBio (and now Ultima) and once for Illumina. It's probably my fault from when I first wrote this doc, but could you please combine them?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -282,6 +345,22 @@ public PathType getPathType() {
     return PathType.VIRTUAL;
   }
 
+  private List<String> getPoolsFromSampleDB(String ampSamplePlate) {
+    List<String> poolNames = new ArrayList<>();
+    try {
+      List<JsonNode> samplePools = apiClient.fetchSampleDB(ampSamplePlate);
+      samplePools.forEach(
+          node -> {
+            if (node.has("libraryPool")) {
```
avarsava, line 359: Is it an issue if a samplePool doesn't have this field?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -80,6 +92,32 @@ private static String fetchNexusApiTokenFile(ObjectNode parameters) {
...
 
+  /**
+   * @param parameters ObjectNode
+   * @return String with base URL for the ULTIMA Sample DB API
+   */
+  private static String fetchSampleDBApiUrl(ObjectNode parameters) {
+    if (parameters.hasNonNull("sampleDBApiAddress")) {
+      return parameters.get("sampleDBApiAddress").asText();
+    } else {
+      log.info("No SampleDB API URL configured for Ultima");
```
avarsava: Are these fields required or optional?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -25,29 +27,54 @@ public class UltimaApiClient {
...
+
+    if (tokenPathSampleDB != null) {
+      try {
+        String tokenSampleDB = Files.readString(Paths.get(tokenPathSampleDB)).trim();
+        HttpHeaders headersSampleDB = new HttpHeaders();
+        headersSampleDB.set(HttpHeaders.AUTHORIZATION, tokenSampleDB);
+        headersSampleDB.set(HttpHeaders.ACCEPT, "application/json");
+        this.headersSampleDB = headersSampleDB;
+      } catch (Exception e) {
+        log.warn("Sample DB token configuration is invalid {}", tokenPathSampleDB, e);
```
avarsava: I think this needs a second `{}`?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -71,15 +98,56 @@ public List<JsonNode> fetchAllRunSummaries() throws IOException {
...
+    String url =
+        String.format("%s/api/v2/lib-pools?samplePlateId=%s", apiUrlSampleDB, samplePlateId);
+
+    HttpEntity<String> entity = new HttpEntity<>(headersSampleDB);
+
+    ResponseEntity<JsonNode> response;
+    try (AutoCloseable latency = sample_db_fetch_time.start()) {
+      response = restTemplate.exchange(url, HttpMethod.GET, entity, JsonNode.class);
+    } catch (Exception e) {
+      api_fetch_failed.inc();
```
avarsava: This metric is now counting failures across two different URLs. Might want to split those up, or label them

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
grapigeau: The LoggerFactory actually treats the last parameter as a throwable so you don't need a placeholder {} for the exception itself

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: Optional (specified that in the appendix now)

avarsava (approved):

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -208,6 +247,30 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+    consumables.add(new Consumable("AMP Wash2", json.path("AMP_Wash2").asText()));
+    consumables.add(
+        new Consumable("AMP Enrichment Bead", json.path("AMP_EnrichmentBead").asText()));
+    consumables.add(new Consumable("Sample Rack", json.path("SampleRack").asText()));
+    consumables.add(new Consumable("Sample Tube", json.path("SampleTube").asText()));
+    consumables.add(
+        new Consumable("Sequencing Cartridge", json.path("SequencingCartridge").asText()));
+    consumables.add(new Consumable("Wash Container", json.path("WashContainer").asText()));
+    consumables.add(new Consumable("Wafer", json.path("Wafer").asText()));
+
```
djcooke, line 267: What are all of these for? The JIRA ticket says "only thing we currently need Sample DB for is Pool linking." We shouldn't inflate the data with stuff we have  no use for.

It's also unclear what a lot of these are. e.g. is "Sample Plate" a name? model? barcode? something else? We don't need to use the same unclear naming that the foreign API does.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
djcooke: Logging this makes it seem like a problem, regardless of the log level, and it doesn't seem helpful even if reworded.

The fetch optional parameter methods could be combined into one method that takes the parameter name, which would eliminate the string duplication too.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -25,29 +27,60 @@ public class UltimaApiClient {
...
+
+    if (tokenPathSampleDB != null) {
+      try {
+        String tokenSampleDB = Files.readString(Paths.get(tokenPathSampleDB)).trim();
+        HttpHeaders headersSampleDB = new HttpHeaders();
+        headersSampleDB.set(HttpHeaders.AUTHORIZATION, tokenSampleDB);
+        headersSampleDB.set(HttpHeaders.ACCEPT, "application/json");
+        this.headersSampleDB = headersSampleDB;
+      } catch (Exception e) {
+        log.warn("Sample DB token configuration is invalid {}", tokenPathSampleDB, e);
```
djcooke: Why is there a try/catch here if not for the nexus token and headers that are set up the exact same way? If the config is invalid, that should be caught before here and the config should show invalid in the UI. A log message doesn't draw the attention that this needs.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -25,29 +27,60 @@ public class UltimaApiClient {
...
+          .register();
+
+  private static final Logger log = LoggerFactory.getLogger(UltimaApiClient.class);
+
+  private final String apiUrlNexus;
   private final RestTemplate restTemplate;
-  private final HttpHeaders headers;
+  private final HttpHeaders headersNexus;
+  private String apiUrlSampleDB = null;
+  private HttpHeaders headersSampleDB = null;
```
djcooke: These can be final too. Just set them to null in the constructor if Sample DB isn't configured.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -56,30 +89,71 @@ public List<JsonNode> fetchAllRunSummaries() throws IOException {
...
+
+    String url =
+        String.format("%s/api/v2/lib-pools?samplePlateId=%s", apiUrlSampleDB, samplePlateId);
+
+    HttpEntity<String> entity = new HttpEntity<>(headersSampleDB);
+
+    ResponseEntity<JsonNode> response;
+    try (AutoCloseable latency_sampledb = sample_db_fetch_time.start()) {
+      response = restTemplate.exchange(url, HttpMethod.GET, entity, JsonNode.class);
+    } catch (Exception e) {
```
djcooke, line 133: Catch specific exception type(s) unless there's a reason you really need to catch all, which should be very rare, and is not the case here.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -56,30 +89,71 @@ public List<JsonNode> fetchAllRunSummaries() throws IOException {
...
+      api_sample_db_fetch_failed.inc();
+      throw new IOException(e);
+    }
+    if (!response.getStatusCode().is2xxSuccessful()) {
+      api_sample_db_fetch_failed.inc();
+      throw new IOException("API request to " + url + " failed: " + response.getStatusCode());
+    }
```
djcooke: Doesn't `RestTemplate#exchange` throw an exception if there's an error status, making this check unnecessary? It probably also handles redirects, in which case there is more than one request and I'm guessing it would return the final status, but I don't know.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -56,30 +89,71 @@ public List<JsonNode> fetchAllRunSummaries() throws IOException {
...
-          "Unexpected JSON response format from API request: " + response.getBody());
+          "Unexpected JSON response format from API request to "
+              + url
+              + " : "
+              + response.getBody());
     }
     return nodes;
   }
+
+  public List<JsonNode> fetchSampleDB(String samplePlateId) throws IOException {
```
djcooke: Half of the parsing/validation is happening here and then you're returning a (non-JSON) list of JSON objects to continue parsing/validation in `DefaultUltima`. Decide where the responsibilities lie and be consistent.

* API client is definitely responsible for pulling data from the API.
* The run processor is definitely responsible for populating the run DTO.
* Who is responsible for parsing and validating the data from the API?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: Added the other ticket to the description, these names are directly from Andrea and match what will be in the SOP

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
grapigeau: _AutoCloseable_ throws "Exception" which is why that's set here

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
grapigeau: Yes you're right, it should throw some sort of HttpError but I like the extra check

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
djcooke: Ah... Gross that the Prometheus library actually returns `AutoCloseable` rather than an implementation that's more specific.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
djcooke: There's no reason we can't use better names that clarify what the values actually are. Also, what is "AMP?" Is it useful to include in all the names?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
djcooke: I would remove it personally, but if you feel the need to keep, it should be more clear that it should be unnecessary. You could move it into the try block and throw an `IllegalStateException`:

```
try (...) {
  response = restTemplate.exchange(url, HttpMethod.GET, entity, JsonNode.class);
  if (!...2xxSuccessful()) {
    throw new IllegalStateException("Unexpected response status: " + status);
  }
} catch (Exception e) ...
```

I did check and it will handle redirects and return the final status, so should always be 2xx if successful, and would have thrown an exception already otherwise.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -346,10 +333,17 @@ public PathType getPathType() {
   }
 
   private List<String> getPoolsFromSampleDB(String ampSamplePlate) {
+    List<JsonNode> pools = new ArrayList<>();
     List<String> poolNames = new ArrayList<>();
     try {
-      List<JsonNode> samplePools = apiClient.fetchSampleDB(ampSamplePlate);
-      samplePools.forEach(
+      JsonNode samplePools = apiClient.fetchSampleDB(ampSamplePlate);
```
djcooke: Is this top-level node really "sample pools" or should this be renamed?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: I think "AMP" is the primer but let me follow up with Andrea

grapigeau: "AMP (short for amplification) is analogous to "clustering" on Illumina. Basically we think of clustering and sequencing as one "run" on Illumina, because those two distinct steps happen without human intervention. On Ultima those two steps happen on different machines and require human intervention and different reagents/consumable " - Andrea
There was a couple changes Sample Rack -> Sequencing Rack and the AMP Sample Tube and the Sample Tube are the same but the rest should stay as is.

djcooke: Sorry, but I strongly disagree. "AMP" can be written "Amplification," and all consumables should indicate what the value actually is. If "AMP Sample Plate" is a barcode, it should be named "Amplification Sample Plate Barcode." I have no idea whether it's a name, model name, model number, serial number, lot number, etc. It is especially ambiguous if it's a barcode because that could be the model, serial, or lot.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -208,6 +234,39 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+        new Consumable(
+            "Amplification Break Container Lot Number", json.path("AMP_BreakContainer").asText()));
+    consumables.add(
+        new Consumable("Amplification Wash 1 Lot Number", json.path("AMP_Wash1").asText()));
+    consumables.add(
+        new Consumable("Amplification Wash 2 Lot Number", json.path("AMP_Wash2").asText()));
+    consumables.add(
+        new Consumable(
+            "Amplification Enrichment Bead Lot Number", json.path("AMP_EnrichmentBead").asText()));
+    consumables.add(new Consumable("Sequencing Rack", json.path("SampleRack").asText()));
```
djcooke: Missed one?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -282,6 +341,29 @@ public PathType getPathType() {
     return PathType.VIRTUAL;
   }
 
+  private List<String> getPoolsFromSampleDB(String ampSamplePlate) {
+    List<JsonNode> pools = new ArrayList<>();
+    List<String> poolNames = new ArrayList<>();
+    try {
+      JsonNode sampleNode = apiClient.fetchSampleDB(ampSamplePlate);
```
djcooke: 1. Still confusing naming. A "sample" shouldn't contain multiple pools. I'm guessing this node represents the run, or wafer, or "sample plate"?
2. Above, you've described `ampSamplePlate` as "Amplification Sample Plate Lot Number." If it's a lot number, that means many sample plates will have the same value, so it doesn't make sense to use as an identifier for the lookup. Are you sure it actually is a lot number?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -282,6 +341,29 @@ public PathType getPathType() {
...
+      if (sampleNode != null && sampleNode.has("pools") && sampleNode.get("pools").isArray()) {
+        sampleNode.get("pools").forEach(pools::add);
```
djcooke: Wait... There can be multiple pools?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
```diff
@@ -82,6 +85,22 @@ public UltimaProcessStatus getUploadStatus() {
...
+    return consumables;
+  }
+
+  public void setPoolNames(List<String> poolNames) {
+    this.poolNames = poolNames;
+  }
+
+  public List<String> getPoolNames() {
+    return poolNames;
+  }
```
djcooke, line 103: `getLaneContents` should be overridden to return the pool per partition (and I think there's only one partition, so should be only one pool?)

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: For AMP Sample Plate, apparently the Barcode # and Lot # are the same

djcooke: Is it unique to the plate or no? If it's unique to the plate, it's a serial number. If it's not unique, then it's a lot number, and I don't understand how that could be used as an identifier here.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
grapigeau: For not there's only one partition but the Ultima format does support multiple pools so I figure it makes the most sense to match what they have

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
grapigeau: OK it IS unique but they mostly use it to determine the affected lot (I had to check again with Andrea but she confirmed)

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
djcooke: `getLaneContents` could check whether `poolNames.size() == 1` and if so return the single pool. If we were putting multiple pools into a partition, we'd combine the pools in MISO so that it became an individual pool.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
djcooke: > they mostly use it to determine the affected lot

How?

I'm skeptical of the other values all being lot numbers, but I guess we can release it, compare, and adjust naming if necessary before starting to actually use them.

grapigeau: The last 3 digits are the serialized and the rest are lot / date / prefix.

I think that makes sense to just know we may need to readjust as necessary.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
grapigeau: Would think then that this should be `String getPoolName` and never return an array?

djcooke: I think leave it as a list, so you can do something with the values and it'll be easier to troubleshoot if there are multiple pools reported from SampleDB. Where does SampleDB even get the pool names? They're not actually on the samplesheet in a way that I'd think it would understand, if at all.

grapigeau: I'm not sure where that's being input - maybe a question for Jeremy. But confirmed the "libraryPool" field matches the pool field we expect in MISO

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
djcooke: bump

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/UltimaNotificationDto.java
djcooke: bump. `getLaneContents` still not implemented

grapigeau: I'm not sure I understand what `getLaneContents` would be doing?
If there's only only pool associated with the run, that single pool will be returned as a one item list as it's implemented now.

djcooke: `getLaneContents` is part of the API. MISO uses this to attach pools to lanes/partitions. Ultima wafers only have one partition, so MISO will consider this partition 1 and try to find the pool for lane 1. We don't want to have to implement different logic to attach pools for Ultima or other platforms when they are all represented in the same way in MISO and it's easy for Run Scanner to accommodate that with the API it already has.

grapigeau: Ah I misunderstood and thought you meant to replace the poolNames, not in addition. Thank you for clarifying

djcooke (approved):

PR #208, opened 2026-06-18
Glt 4806.q30 metric ultima

grapigeau:
JIRA Ticket: https://jira.oicr.on.ca/browse/GLT-4806

- [x] Updates release notes


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
```diff
@@ -120,6 +125,43 @@ public List<JsonNode> fetchAllRunSummaries() throws IOException {
...
+                + " failed, unexpected response status: "
+                + response.getStatusCode());
+      }
+    } catch (Exception e) {
+      api_nexus_fetch_failed.inc();
+      throw new IOException(e);
+    }
+    if (response.getBody() != null) {
+      if (response.getBody().isArray()) {
+        return response.getBody().get(0); // API returns a list of 1
```
avarsava, line 158: Will it always be an array of 1, going into the future? is it worth testing the array length and throwing if it's >1?

RELEASE_NOTES.md
```diff
@@ -3,6 +3,8 @@
 Changes:
 * Add Sample DB support for Ultima
 * Add consumables to Ultima Runs
+* Add "," to Ultima metrics (1000 -> 1,000) and reduce Total Beads to 6 decimal precision
```
avarsava, line 6: Doesn't this make the value less easily machine-processed?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaApiClient.java
grapigeau: I don't see it being more than 1 item but I added the check in jic ultima does something unexpected

RELEASE_NOTES.md
grapigeau: Theoretically, but right now it's all humans looking at these values in MISO and so this was requested (also matches Illumina metrics format)

avarsava: If it's wanted for a prettier display, I think that should happen in the front end code in MISO

grapigeau: MISO just views (this type of) metric as a "table" which is just key/value pairs and doesn't do any modifications to the table contents. (Not sure how set in stone that approach is)

djcooke: The whole metrics blob is an escaped JSON string within the JSON output of Run Scanner. That escaped JSON string gets stored in the MISO database in the same format. MISO just reads and displays this table data as is. Other systems read the JSON string from the MISO database and parse the numbers out of it. None of this is great, but that's how it works currently.

djcooke (approved):

RELEASE_NOTES.md
avarsava: If this value is getting passed to MISO and then other systems are also parsing it from the database, that says even more to me that this isn't the place for this kind of change. I don't believe Jackson will parse a number containing `,` out of the box? So if Jackson offers a parsing option for it, downstream would need to know to use it just for this value.

avarsava (approved):

PR #214, opened 2026-07-03
Glt 4413.google bucket support

grapigeau:
JIRA Ticket: https://jira.oicr.on.ca/browse/GLT-4813

- [x] Updates release notes

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,11 @@
 # Unreleased
 
 Changes:
+* Add in support for Google Cloud Storage for Ultima
+  * only supported filesystem for Ultima
+* Pipeline Runs added to Ultima
+
+Changes:
 * Handle Ultima container model parsing according to their naming conventions
 * Scan sequencer directories round-robin (instead of finishing one before starting the next)
 
@@ -21,16 +26,12 @@ Fixes:
 # 2.6.0
 
 Changes:
-* Add in support for Google Cloud Storage for Ultima
-
-Changes:
 * Add Sample DB support for Ultima
 * Add consumables to Ultima Runs
 * Add "," to Ultima metrics (1000 -> 1,000) and reduce Total Beads to 6 decimal precision
 * Add "Control Sample Bases > Q30 %" Ultima metric
 * Update Illumina run directory scanning to be completed in reverse order by run directory name (approximately reverse chronological order)
 * Update Ultima run directory scanning to be completed in reverse order by run ID (approximately reverse chronological order)
-
 
 # 2.5.1
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -21,6 +21,9 @@ Fixes:
 # 2.6.0
 
 Changes:
+* Add in support for Google Cloud Storage for Ultima
+
+Changes:
 * Add Sample DB support for Ultima
 * Add consumables to Ultima Runs
 * Add "," to Ultima metrics (1000 -> 1,000) and reduce Total Beads to 6 decimal precision
--- a/docs/appendices.md
+++ b/docs/appendices.md
@@ -106,6 +106,10 @@
                 <td>sampleDBApiTokenFile (optional)</td>
                 <td>The location of a file containing a Sample DB Auth Token</td>
             </tr>
+            <tr>
+                <td>googleServiceAccount (required)</td>
+                <td>The email address of the Google service account to impersonate when accessing the GCS bucket containing run output (Ex. runscanner@project.iam.gserviceaccount.com)</td>
+            </tr>
         </table></td>
     </tr>
 </table>
@@ -380,6 +384,13 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td></td>
         <td></td>
         <td></td>
+    </tr>
+    <tr>
+        <td>workflowRuns</td>
+        <td></td>
+        <td></td>
+        <td></td>
+        <td>✓</td>
     </tr>
 </table>
 
--- a/docs/appendices.md
+++ b/docs/appendices.md
@@ -373,22 +373,15 @@ esp bc they might vary in meaning b/w PromION and MinION -->
     </tr>
     <tr>
         <td>analysisExpected</td>
-        <td>✓</td>
         <td></td>
+        <td>✓</td>
         <td></td>
         <td></td>
     </tr>
     <tr>
         <td>pipelineRuns</td>
+        <td></td>
         <td>✓</td>
-        <td></td>
-        <td></td>
-        <td></td>
-    </tr>
-    <tr>
-        <td>workflowRuns</td>
-        <td></td>
-        <td></td>
         <td></td>
         <td>✓</td>
     </tr>
--- a/docs/appendices.md
+++ b/docs/appendices.md
@@ -107,8 +107,8 @@
                 <td>The location of a file containing a Sample DB Auth Token</td>
             </tr>
             <tr>
-                <td>googleServiceAccount (required)</td>
-                <td>The email address of the Google service account to impersonate when accessing the GCS bucket containing run output (Ex. runscanner@project.iam.gserviceaccount.com)</td>
+                <td>googleCredentialsFile (required)</td>
+                <td>The location of a Google service account JSON key file, used to authenticate when accessing the GCS bucket containing run output</td>
             </tr>
         </table></td>
     </tr>
--- a/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaAnalysisUnit.java
+++ b/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaAnalysisUnit.java
@@ -5,7 +5,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
 
-// Corresponds to one Barcode Folder in GCS: {runId}-{library}-{barcode}/
+// Corresponds to one Barcode Folder: {runId}-{library}-{barcode}/
 public class UltimaAnalysisUnit {
 
   private String barcode;
--- a/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaAnalysisUnit.java
+++ b/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaAnalysisUnit.java
@@ -1,0 +1,65 @@
+package ca.on.oicr.gsi.runscanner.dto.ultima;
+
+import ca.on.oicr.gsi.runscanner.dto.AnalysisFile;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+// Corresponds to one Barcode Folder in GCS: {runId}-{library}-{barcode}/
+public class UltimaAnalysisUnit {
+
+  private String barcode;
+  private String library;
+  private List<AnalysisFile> files = new LinkedList<>();
+
+  public String getBarcode() {
+    return barcode;
+  }
+
+  public void setBarcode(String barcode) {
+    this.barcode = barcode;
+  }
+
+  public String getLibrary() {
+    return library;
+  }
+
+  public void setLibrary(String library) {
+    this.library = library;
+  }
+
+  public List<AnalysisFile> getFiles() {
+    return files;
+  }
+
+  public void setFiles(List<AnalysisFile> files) {
+    this.files = files;
+  }
+
+  @Override
+  public String toString() {
+    return "UltimaAnalysisUnit [barcode="
+        + barcode
+        + ", library="
+        + library
+        + ", files="
+        + files
+        + "]";
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) return true;
+    if (obj == null) return false;
+    if (getClass() != obj.getClass()) return false;
+    UltimaAnalysisUnit other = (UltimaAnalysisUnit) obj;
+    return Objects.equals(barcode, other.barcode)
+        && Objects.equals(library, other.library)
+        && Objects.equals(files, other.files);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(barcode, library, files);
+  }
+}
--- a/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaWorkflowRun.java
+++ b/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaWorkflowRun.java
@@ -1,6 +1,7 @@
 package ca.on.oicr.gsi.runscanner.dto.ultima;
 
 import ca.on.oicr.gsi.runscanner.dto.WorkflowRun;
+import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Objects;
@@ -9,8 +10,8 @@ public class UltimaWorkflowRun extends WorkflowRun {
 
   private List<UltimaAnalysisUnit> analysisOutputs = new LinkedList<>();
 
-  public UltimaWorkflowRun() {
-    super("ultimaWorkflowRun");
+  public UltimaWorkflowRun(@JsonProperty("workflowName") String workflowName) {
+    super(workflowName);
   }
 
   public List<UltimaAnalysisUnit> getAnalysisOutputs() {
--- a/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaWorkflowRun.java
+++ b/dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaWorkflowRun.java
@@ -1,0 +1,42 @@
+package ca.on.oicr.gsi.runscanner.dto.ultima;
+
+import ca.on.oicr.gsi.runscanner.dto.WorkflowRun;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+public class UltimaWorkflowRun extends WorkflowRun {
+
+  private List<UltimaAnalysisUnit> analysisOutputs = new LinkedList<>();
+
+  public UltimaWorkflowRun() {
+    super("initialCramGeneration");
+  }
+
+  public List<UltimaAnalysisUnit> getAnalysisOutputs() {
+    return analysisOutputs;
+  }
+
+  public void setAnalysisOutputs(List<UltimaAnalysisUnit> analysisOutputs) {
+    this.analysisOutputs = analysisOutputs;
+  }
+
+  @Override
+  public String toString() {
+    return super.toString() + ", UltimaWorkflowRun [analysisOutputs=" + analysisOutputs + "]";
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) return true;
+    if (!super.equals(obj)) return false;
+    if (getClass() != obj.getClass()) return false;
+    UltimaWorkflowRun other = (UltimaWorkflowRun) obj;
+    return Objects.equals(this.analysisOutputs, other.analysisOutputs);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(super.hashCode(), analysisOutputs);
+  }
+}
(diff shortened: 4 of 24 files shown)


docs/appendices.md
```diff
@@ -381,6 +385,13 @@ esp bc they might vary in meaning b/w PromION and MinION -->
         <td></td>
         <td></td>
     </tr>
+    <tr>
+        <td>workflowRuns</td>
```
avarsava: pipelineRuns has workflowRuns embedded in it - could it be consistent with other runs? 

Image

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -266,9 +295,101 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+        String folderName = extractFolderName(folder.getName());
+
+        // Folder name: {runId}-{library}-{barcode}
+        int underscoreIdx = folderName.indexOf('-');
+        if (underscoreIdx < 0) continue;
+        String libBarcode = folderName.substring(underscoreIdx + 1);
+        int lastDash = libBarcode.lastIndexOf('-');
+        if (lastDash < 0) continue;
+        String library = libBarcode.substring(0, lastDash);
+        String barcode = libBarcode.substring(lastDash + 1);
```
avarsava:
```suggestion
        String[] folderParts = folderName.split("-");
        if (folderParts.length != 3) continue;
        String library = folderParts[1];
        String barcode = folderParts[2];
```

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -266,9 +295,101 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+    List<AnalysisFile> files = new ArrayList<>();
+    try {
+      for (Blob blob : googleBucketClient.ls(barcodeFolderPath)) {
+        if (blob.isDirectory()) continue;
+        String blobName = blob.getName();
+        // Use CramAnalysisFile for .cram files; MetadataAnalysisFile for everything else
+        AnalysisFile file =
+            blobName.endsWith(".cram") ? new CramAnalysisFile() : new MetadataAnalysisFile();
+        // Path stored as "bucket/object"
+        // TODO do we want to be clear that it's GCS aka gs://bucket/object by including "gs://" in
```
avarsava: Yes please, other paths have `file:/` for contrast

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -365,6 +486,33 @@ private List<String> getPoolsFromSampleDB(String ampSamplePlate) {
...
+    }
+  }
+
+  private static String extractFolderName(String blobName) {
+    // GCS dir blobs have a trailing '/'. Strip it, then take everything after the last '/' to get
+    // the folder name
+    String stripped =
+        blobName.endsWith("/") ? blobName.substring(0, blobName.length() - 1) : blobName;
+    int lastSlash = stripped.lastIndexOf('/');
+    return lastSlash >= 0 ? stripped.substring(lastSlash + 1) : stripped;
```
avarsava:
```suggestion
    String[] blobNameParts = blobName.split("/");
    return blobNameParts[blobNameParts.length-1]
```
split() drops trailing empty strings, so this is tolerant of having a / at the end or not: https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html#split(java.lang.String)

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaGoogleBucketClient.java
```diff
@@ -0,0 +1,66 @@
...
+    return blobs;
+  }
+
+  private static String[] parseGcsPath(String gcsPath) {
+    // Paths are passed as "bucket/runfolder". Split on the first '/' to separate the bucket name
+    int slashIdx = gcsPath.indexOf('/');
+    if (slashIdx < 0) {
+      return new String[] {gcsPath, ""};
+    }
+    return new String[] {gcsPath.substring(0, slashIdx), gcsPath.substring(slashIdx + 1)};
```
avarsava:
```suggestion
    // Paths are passed as "bucket/runfolder". Split on the first '/' to separate the bucket name
    return gcsPath.split("/");
```

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -350,10 +350,14 @@ private List<UltimaAnalysisUnit> buildAnalysisUnits(String runId, String runFold
...
-        String library = folderParts[1];
-        String barcode = folderParts[2];
+        // Split on the first and last '-', library names can contain dashes
+        int underscoreIdx = folderName.indexOf('-');
+        if (underscoreIdx < 0) continue;
+        String libBarcode = folderName.substring(underscoreIdx + 1);
+        int lastDash = libBarcode.lastIndexOf('-');
+        if (lastDash < 0) continue;
+        String library = libBarcode.substring(0, lastDash);
+        String barcode = libBarcode.substring(lastDash + 1);
```
avarsava: Commit 'fixup and better comments' undid this change?

avarsava: Maybe didn't pull the applied suggestion commits from the remote

grapigeau: So I realized why I didn't use your simpler solution off the bat. library names can contain dashes so can't split on every "-" (I added a comment that hopefully helps?)

avarsava (approved):

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaAnalysisUnit.java
```diff
@@ -0,0 +1,65 @@
+package ca.on.oicr.gsi.runscanner.dto.ultima;
+
+import ca.on.oicr.gsi.runscanner.dto.AnalysisFile;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+// Corresponds to one Barcode Folder in GCS: {runId}-{library}-{barcode}/
```
djcooke: Is GCS the only storage option for Ultima analysis? Shouldn't be documented this way if not.

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaWorkflowRun.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+
+public class UltimaWorkflowRun extends WorkflowRun {
+
+  private List<UltimaAnalysisUnit> analysisOutputs = new LinkedList<>();
+
+  public UltimaWorkflowRun() {
+    super("initialCramGeneration");
```
djcooke: Is this not intended to support other workflows? Rename to `UltimaCramGenerationWorkflow` or something similar if that's all that it supports.

RELEASE_NOTES.md
```diff
@@ -1,15 +1,15 @@
...
-
 # 2.6.1
 
 Fixes:
 * "Control Sample Bases > Q30 %" to come from subtable in metrics
 
 # 2.6.0
 
+Changes:
+* Add in support for Google Cloud Storage for Ultima
```
djcooke: * It looks like GCS is a requirement now and no other storage options are supported. That should be stated more plainly here.
* The changelog should mention how the output has changed (pipeline runs added).

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/UltimaGoogleBucketClient.java
```diff
@@ -0,0 +1,67 @@
...
+  /** Mimics: gcloud storage ls --impersonate-service-account {serviceAccount} {gcsPath} */
+  public List<Blob> ls(String gcsPath) {
+    String[] parts = parseGcsPath(gcsPath);
+    String bucket = parts[0];
+    String runPath = parts[1];
+    // GCS prefix-based listing treats a trailing '/' as a directory boundary. Without it, a prefix
+    // of "seq" would also match "seq2/", "seq-old/", etc.
+    if (!runPath.isEmpty() && !runPath.endsWith("/")) {
+      runPath = runPath + "/";
+    }
```
djcooke: Should this be part of `parseGcsPath`? If that were called somewhere else, it would need the same treatment, right?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -266,9 +298,106 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
 
     dto.setConsumables(consumables);
 
+    String runFolder = runFolderCache.get(runId);
+    dto.setSequencerFolderPath(runFolder != null ? runFolder : runDirectory.getPath());
+    if (runFolder != null) {
+      UltimaWorkflowRun workflowRun = buildWorkflowRun(runId, runFolder, json, uploadStatus);
```
djcooke: What happens here if the run is still sequencing? Is `runFolder` null? Please make note of this if so.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -266,9 +298,106 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+    UltimaWorkflowRun workflowRun = new UltimaWorkflowRun();
+
+    // Analysis_Start_Time uses the same "yyyy-MM-dd HH:mm:ss" format as startdatetime
+    String analysisStartStr = json.path("Analysis_Start_Time").asText("");
+    if (!analysisStartStr.isBlank()) {
+      try {
+        LocalDateTime ldt = LocalDateTime.parse(analysisStartStr, NEXUS_DATE_FORMATTER);
+        workflowRun.setStartTime(extractTime(ldt, json.path("timezone").asText()));
+      } catch (Exception e) {
+        log.warn("Could not parse Analysis_Start_Time '{}' for run {}", analysisStartStr, runId);
```
djcooke: Catch specific exception type

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -266,9 +298,106 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+        file.setSize(blob.getSize());
+        if (blob.getCreateTimeOffsetDateTime() != null) {
+          file.setCreatedTime(blob.getCreateTimeOffsetDateTime().toInstant());
+        }
+        if (blob.getUpdateTimeOffsetDateTime() != null) {
+          file.setModifiedTime(blob.getUpdateTimeOffsetDateTime().toInstant());
+        }
+        files.add(file);
+      }
+    } catch (Exception e) {
```
djcooke: Catch specific exception type

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -365,6 +494,31 @@ private List<String> getPoolsFromSampleDB(String ampSamplePlate) {
...
+        if (!blob.isDirectory()) continue;
+        String folderName = extractFolderName(blob.getName());
+        // Run folders are named {runId}-{YYYYMMDD}_{suffix}
+        int dashIdx = folderName.indexOf('-');
+        if (dashIdx < 0)
+          continue; // only include folders with run directory formatting used by the Ultima machine
+        String runId = folderName.substring(0, dashIdx);
+        runFolderCache.put(runId, blob.getBucket() + "/" + blob.getName());
+      }
+    } catch (Exception e) {
```
djcooke: Catch specific exception type

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -266,9 +298,106 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
 
     dto.setConsumables(consumables);
 
+    String runFolder = runFolderCache.get(runId);
+    dto.setSequencerFolderPath(runFolder != null ? runFolder : runDirectory.getPath());
+    if (runFolder != null) {
+      UltimaWorkflowRun workflowRun = buildWorkflowRun(runId, runFolder, json, uploadStatus);
+      // Ultima only ever has one attempt per runId.
+      UltimaPipelineRun pipelineRun = new UltimaPipelineRun(1);
```
djcooke: There are a bunch of places below where you're catching exceptions and logging warnings and just continuing to populate the pipeline run data. This can result in something weird, like reporting a complete pipeline, but having no analysis files, and that won't be obvious to a consumer because they're not going to check the logs along with the JSON output.

Creating a pipeline run should be separated into its own method and if any part of it fails, the pipeline run should either not be included, or should have some kind of unknown/error status.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -266,9 +298,106 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
...
+        String barcode = libBarcode.substring(lastDash + 1);
+
+        String barcodeFolderPath = folder.getBucket() + "/" + folder.getName();
+        UltimaAnalysisUnit unit = new UltimaAnalysisUnit();
+        unit.setBarcode(barcode);
+        unit.setLibrary(library);
+        unit.setFiles(buildAnalysisFiles(barcodeFolderPath));
+        units.add(unit);
+      }
+    } catch (Exception e) {
```
djcooke: Catch specific exception type

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -198,6 +201,8 @@ public NotificationDto process(File runDirectory, TimeZone tz) throws IOExceptio
     double passFilterPercent = (numBeads > 0) ? (numReadPass / numBeads) * 100 : 0;
     double controlBasesQ30 = getControlQ30Bases(runId);
 
+    int test1 = json.path("pf_output_reads").asInt(0);
+    String test2 = json.path("pf_output_reads").toString();
```
djcooke: ?

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaAnalysisUnit.java
djcooke: For clarification: I meant to ask whether GCS is the only storage option that _Ultima_ supports (not Run Scanner)? If Run Scanner may be expanded in the future to support other storage options, then it doesn't make sense to document "UltimaAnalysisUnit" as being specific to GCS.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -415,7 +424,7 @@ private Instant extractTime(LocalDateTime datetime, String timeZone) {
     try {
       ZoneId zone = TimeZone.getTimeZone(timeZone).toZoneId();
       return datetime.atZone(zone).toInstant();
-    } catch (Exception e) {
+    } catch (DateTimeException e) {
```
djcooke: This error is still being ignored when building a pipeline run

docs/appendices.md
```diff
@@ -107,8 +107,8 @@
                 <td>The location of a file containing a Sample DB Auth Token</td>
             </tr>
             <tr>
-                <td>googleServiceAccount (required)</td>
-                <td>The email address of the Google service account to impersonate when accessing the GCS bucket containing run output (Ex. runscanner@project.iam.gserviceaccount.com)</td>
+                <td>googleCredentialsFile (required)</td>
+                <td>The location of a Google service account JSON key file, used to authenticate when accessing the GCS bucket containing run output</td>
```
avarsava, line 111: Is this key file something Google provides? or should the expected format be provided here?

avarsava (approved):

docs/appendices.md
grapigeau: Google generates this

dto/src/main/java/ca/on/oicr/gsi/runscanner/dto/ultima/UltimaWorkflowRun.java
```diff
@@ -10,7 +10,7 @@ public class UltimaWorkflowRun extends WorkflowRun {
   private List<UltimaAnalysisUnit> analysisOutputs = new LinkedList<>();
 
   public UltimaWorkflowRun() {
-    super("initialCramGeneration");
+    super("ultimaWorkflowRun");
```
djcooke: The workflow name needs to be specific. If/when there are more workflows, you need to be able to determine which is the cram generation workflow, etc. My previous comment was that the workflow name was hard-coded, making it so the "UltimaWorkflowRun" class only represented a specific workflow, while the class name suggested a larger scope.

Either:

1. Hard-code the workflow name as it was before, and rename the class to match, or
2. Take the workflow name as a constructor parameter and leave the class name as is

Note also that the existing example workflow name already in Run SCanner is "BCLConvert" so I think the proper workflow name should be used, rather than a camelCase representation.

grapigeau: I think all workflows will export their files here and we'll handle that as one big "Ultima output" dump. So we want to keep this workflow name generic (at least for now).
 The contents are variable, depending on how the run is set up.  for example, em-seq runs have additional files, ppmseq vs non-ppmseq differ.

djcooke: This is a workflow run. It should represent a run of one specific workflow. You're talking about a pipeline run, which may include multiple workflow runs.

grapigeau: We don't really know what workflow is being run though? Even Larry just suggested UltimaInstrumentWorkflow to describe what's happening on instrument and being output to a barcode dir

djcooke: What does AnalysisRecipe look like? Can the cram files be associated with a "cram generation" workflow? What are the other "metadata files" that are being included?

grapigeau: AnalysisRecipe looks something like: `WGS1;SDB_EC.yaml` or for calibration `Other;SDB_EC_TT_APP.yaml`
The other metadata files can vary but always include a .csv and .json with some metrics as well as trimming metadata (failures + stats).
I think the idea was we didn't want to parse each of the file names and extensions at this point to group them (but may want to later on, there's been some back and forth).

djcooke: You are already identifying the cram files though, so maybe best to at least group them into a specific workflow run? The other files, if we don't have a specific use for them, maybe it's best not to include them for now?

If we implement it like this, downstream things will be built to depend on it and it will become more difficult to change. If there end up being multiple workflows and we need to know the files for each of them, the mess will only grow.

djcooke: And what do the yaml files referenced in the AnalysisRecipe contain?

grapigeau: I do not know - Larry is off today but I'll look into it

djcooke (approved):

PR #215, opened 2026-07-03
return correct error response when deleting unreadable run

grapigeau:
JIRA Ticket: https://jira.oicr.on.ca/browse/GP-5610

- [x] Updates release notes

### DIFF
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/UserInterfaceController.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/UserInterfaceController.java
@@ -38,6 +38,7 @@ import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
 
 /** Front-end status monitoring for run scanner */
 @Controller
@@ -245,7 +246,7 @@ public class UserInterfaceController {
     }
   }
 
-  @GetMapping("/error")
+  @RequestMapping("/error")
   public void showError(HttpServletRequest request, HttpServletResponse response)
       throws IOException {
     response.setContentType(CONTENT_TYPE);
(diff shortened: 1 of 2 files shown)


scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/UserInterfaceController.java
```diff
@@ -245,7 +246,7 @@ protected void emitCore(SectionRenderer renderer) throws XMLStreamException {
     }
   }
 
-  @GetMapping("/error")
+  @RequestMapping("/error")
```
avarsava, line 249: I'm not well versed in Spring - how does this help? I thought the problem was in RestResponseController?

grapigeau: So basically a ResourceNotFoundException is thrown and Spring should translate that into a 404 however Tomcat is doing something in between using /error (I'm not 100% confident about what this step is doing). The UserInterfaceController then gets a DELETE /error and then Spring returns 405 method not allowed b/c only GET is allowed at that mapping

avarsava: Is it tomcat or is it spring? looks like we have an `error page` set here: https://github.com/miso-lims/runscanner/blob/master/scanner/src/main/webapp/WEB-INF/web.xml#L62

I don't know how useful this setting is. Weird that it carries the method through.

avarsava (approved):

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/UserInterfaceController.java
grapigeau: Yeah the web.xml error-page "redispatches" the error with Tomcat, so I guess it is both Tomcat and Spring.
`If an error-page element in the deployment descriptor does not contain an exception-type or an error-code element, the error page is a default error page.` so I also don't know how useful this is https://jakarta.ee/specifications/servlet/6.0/jakarta-servlet-spec-6.0#deployment-descriptor

callunity (approved):

PR #219, opened 2026-07-14
Gp 5578.round robin scanning

grapigeau:
JIRA Ticket: https://jira.oicr.on.ca/browse/GP-5578
- x ] Updates release notes

### DIFF
--- a/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
+++ b/scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
@@ -401,7 +401,7 @@ public class Scheduler {
    */
   protected static Stream<Pair<File, Configuration>> roundRobin(
       List<List<Pair<File, Configuration>>> sequencerRuns) {
-    List<Pair<File, Configuration>> mergedRuns = new ArrayList<>();
+    List<Pair<File, Configuration>> mergedRuns = new ArrayList<>(sequencerRuns.stream().mapToInt(List::size).sum());
     int longestRunCount = sequencerRuns.stream().mapToInt(List::size).max().orElse(0);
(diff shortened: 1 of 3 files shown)


djcooke (approved):

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
```diff
@@ -392,6 +388,27 @@ public boolean invalidate(String runName) {
...
+  /**
+   * Merge several lists into one, taking one element from each list in turn (round-robin) rather
+   * than exhausting one list before moving to the next.
+   *
+   * <p>This is used so that, when there is a backlog of runs to process, no single sequencer's runs
+   * monopolize the processing queue ahead of another's.
+   */
+  protected static Stream<Pair<File, Configuration>> roundRobin(
+      List<List<Pair<File, Configuration>>> sequencerRuns) {
+    List<Pair<File, Configuration>> mergedRuns = new ArrayList<>();
```
avarsava: @djcooke might know more about this than me - but would


```suggestion
    List<Pair<File, Configuration>> mergedRuns = new ArrayList<>(sequencerRuns.stream().mapToInt(List::size).sum());
```

end up with any sort of performance improvement? since there wouldn't need to be any pause to resize the underlying array?

avarsava (approved):

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/Scheduler.java
djcooke: I think it's very likely that would be more efficient. This whole method could be much more optimized, but it's not running often enough to make a noticeable difference.

PR #221, opened 2026-07-22
handle error when config not setup

grapigeau:
JIRA Ticket:

- [x] Updates release notes


avarsava (approved):

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/DefaultUltima.java
```diff
@@ -56,7 +56,7 @@ public static DefaultUltima create(Builder builder, ObjectNode parameters) {
           fetchGoogleCredentialsFile(parameters),
           fetchOptionalParameter(parameters, "sampleDBApiAddress"),
           fetchOptionalParameter(parameters, "sampleDBApiTokenFile"));
-    } catch (IOException e) {
+    } catch (IOException | NullPointerException e) {
```
djcooke: The NPE is a programming error - it should be avoided rather than caught

djcooke (approved):

PR #222, opened 2026-07-28
Google bucket or file support

grapigeau:
JIRA Ticket:

- [x] Updates release notes


docs/appendices.md
```diff
@@ -107,7 +107,11 @@
                 <td>The location of a file containing a Sample DB Auth Token</td>
             </tr>
             <tr>
-                <td>googleCredentialsFile (required)</td>
+                <td>useGoogleBucket (optional, default true)</td>
+                <td>Whether the <code>path</code> above points to a Google Cloud Storage bucket (<code>true</code>, e.g. <code>bucket-name/some/prefix</code>) or a local filesystem directory (<code>false</code>, e.g. <code>/absolute/path/to/sequencer/output</code>). Run folder/barcode folder/file layout underneath is identical either way.</td>
```
avarsava: I would clarify this to specify the "path" which specifies the "Absolute path to sequencer output". I did think this meant the 'sampleDBApiTokenFile' at first.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/ultima/LocalStorageClient.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public long getSize() {
+      return size;
+    }
+
+    @Override
+    public String getCrc32Checksum() {
+      // TODO No local equivalent to GCS's CRC32C implemented currently, maybe needed?
```
djcooke: This is needed for analysis files, right?

docs/appendices.md
```diff
@@ -107,7 +107,11 @@
                 <td>The location of a file containing a Sample DB Auth Token</td>
             </tr>
             <tr>
-                <td>googleCredentialsFile (required)</td>
+                <td>useGoogleBucket (optional, default true)</td>
```
avarsava, line 110: Don't GCS addresses start with `gs://` ? Could the processor decide what to do based on that?
And is there any possibility that another type of URL might be used in the future?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/ultima/LocalStorageClient.java
```diff
@@ -0,0 +1,101 @@
...
+      return null;
+    }
+
+    @Override
+    public Instant getCreatedTime() {
+      // Most local filesystems don't expose a reliable creation time; reuse modified time
```
djcooke: You might as well try before falling back to the modification time

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/ultima/LocalStorageClient.java
```diff
@@ -0,0 +1,101 @@
...
+
+  /** Mimics: ls {path} */
+  @Override
+  public List<UltimaStorageEntry> ls(String path) throws IOException {
+    Path dir = Paths.get(path);
+    List<UltimaStorageEntry> entries = new ArrayList<>();
+    try (Stream<Path> children = Files.list(dir)) {
+      children.forEach(child -> entries.add(new LocalStorageEntry(child)));
+    } catch (NoSuchFileException | NotDirectoryException e) {
+      // Mirrors GCS listing semantics: a path with no matching contents yields an empty list not an
```
avarsava: Wouldn't you want an error if the user configured an invalid path?

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/ultima/LocalStorageClient.java
avarsava: Depends what this is being used for? CRC32 is not meant for checking between files, only the same file being transferred somewhere. So if local files use a slightly different algorithm than gcs files I think it's okay if they won't be compared to each other. 

Here is an example of how to calculate a crc32: https://github.com/oicr-gsi/vidarr/blob/master/vidarr-core/src/main/java/ca/on/oicr/gsi/vidarr/core/LocalOutputProvisioner.java#L33

avarsava: If the intention is to load these into Vidarr, they will need some kind of checksum for that

docs/appendices.md
grapigeau: There's no future plans for another type but we're also not using the gs:// in the config file (it needs to be excluded in all requests to GCS)

avarsava (approved):

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/ultima/LocalStorageClient.java
```diff
@@ -83,14 +84,26 @@ public long getSize() {
 
     @Override
     public String getCrc32Checksum() {
-      // TODO No local equivalent to GCS's CRC32C implemented currently, maybe needed?
-      return null;
+      // Note: this is a plain CRC32, not GCS's CRC32C, and is computed on demand since it's only
+      // needed for a subset of files (e.g. .cram files).
```
djcooke: "On demand" for Run Scanner means when serializing, which will happen for all files. Since the storage entry will be stored in memory and serialized many times, it would be much more efficient to compute the checksum once and keep the result in a field. Otherwise, all of the files need to be read every time their runs are serialized.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/ultima/LocalStorageClient.java
```diff
@@ -37,22 +33,27 @@ private static class LocalStorageEntry implements UltimaStorageEntry {
...
       if (!directory) {
         try {
-          fileSize = Files.size(path);
-          lastModified = Files.getLastModifiedTime(path).toInstant();
+          BasicFileAttributes attributes = Files.readAttributes(path, BasicFileAttributes.class);
+          fileSize = attributes.size();
+          lastModified = attributes.lastModifiedTime().toInstant();
+          created = attributes.creationTime().toInstant();
         } catch (IOException e) {
           log.warn("Could not read file attributes for {}", path, e);
```
djcooke: This exception means you can't generate a valid entry, so it shouldn't be caught here

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/ultima/LocalStorageClient.java
```diff
@@ -84,20 +84,23 @@ public long getSize() {
 
     @Override
     public String getCrc32Checksum() {
-      // Note: this is a plain CRC32, not GCS's CRC32C, and is computed on demand since it's only
-      // needed for a subset of files (e.g. .cram files).
+      // Note: this is a plain CRC32, not GCS's CRC32C.
+      if (crc32Computed) {
+        return crc32Checksum;
+      }
```
djcooke: Lazy init means that this will be calculated the first time the run is retrieved from Run Scanner, rather than during scanning. Better to do the file IO during scanning so it doesn't affect response times.

scanner/src/main/java/ca/on/oicr/gsi/runscanner/scanner/processor/ultima/LocalStorageClient.java
```diff
@@ -49,11 +48,26 @@ private static class LocalStorageEntry implements UltimaStorageEntry {
...
+    private static String computeCrc32Checksum(Path path) {
+      // Note: this is a plain CRC32, not GCS's CRC32C.
+      try {
+        byte[] data = Files.readAllBytes(path);
+        CRC32 crc32 = new CRC32();
+        crc32.update(data);
+        return Long.toString(crc32.getValue());
+      } catch (IOException e) {
+        log.warn("Could not calculate CRC32 checksum for {}", path, e);
+        return null;
```
djcooke: The data is invalid without the checksum, right? The exception shouldn't be caught here.

